PR #1, opened 2019-11-29
React dashboard

jacekkolasa:
Those are the changes which were added originally in https://github.com/netdata/netdata/pull/6516. I have just prepend it with with initial commit, which adds few old-codebase files, so the following commits could be added on top.

As i've mentioned few times, please note that some parts of the new codebase can use some work, ie. refractoring to match our codestyle, like functional, declarative programming - but i've added `todo` comments in those places. 

After this PR, i'll create 3 more PRs which have already code in other branches, for easier evaluation.

Because this PR is so big, perhaps we could have your eventual notes addressed in separate PR?


src/domains/global/selectors.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { AppStateT } from "store/app-state"
+
+import { GetKeyArguments, getKeyForCommonColorsState } from "./reducer"
+import { storeKey } from "./constants"
+
+export const createSelectAssignedColors = (args: GetKeyArguments) => (state: AppStateT) => {
+  const keyName = getKeyForCommonColorsState(args)
+  const substate = state[storeKey].commonColorsKeys[keyName]
+  return substate && substate.assigned
```
jacekkolasa: @Azein re: your comment from previous CR:
> For variable-dependant selection and &&-based protection just the R.path could be used, I think.

I have tried R.path but the typing doesn't work that well in that scenario, how about if we use optional chaining here? (that feature is not yet added to the codebase)

Azein: I'm all for it, we are adopting it in cloud-fe currently, works nice.

PR #10, opened 2020-01-07
7351-global-pan-zoom-mainjs

jacekkolasa:
hooked
- global pan and zoom
- global highlight
to main.js. That means:
- url gets updated on redux state change
- redux state is initialized based on initial url

### DIFF
--- a/src/domains/chart/components/portals.tsx
+++ b/src/domains/chart/components/portals.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { memo } from "react"
 import { createPortal } from "react-dom"
 
 import { getAttributes } from "../utils/transformDataAttributes"
@@ -7,7 +7,7 @@ import { DisableOutOfView } from "./disable-out-of-view"
 
 const getNodesArray = () => Array.from(document.querySelectorAll("[data-netdata]"))
 
-export const Portals = () => {
+export const Portals = memo(() => {
   const nodes = getNodesArray()
   return (
     <>
@@ -32,4 +32,4 @@ export const Portals = () => {
       })}
     </>
   )
-}
+})
--- a/src/domains/chart/utils/formatters.ts
+++ b/src/domains/chart/utils/formatters.ts
@@ -129,7 +129,7 @@ export const useFormatters = ({
   } = attributes
 
   const legendFormatValueDecimalsFromMinMax = useCallback((newMin: number, newMax: number) => {
-    if (safeEqualCheck(newMin, min) && safeEqualCheck(newMax, max)) {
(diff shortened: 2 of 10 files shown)


src/domains/chart/components/portals.tsx
```diff
@@ -7,7 +7,7 @@ import { DisableOutOfView } from "./disable-out-of-view"
 
 const getNodesArray = () => Array.from(document.querySelectorAll("[data-netdata]"))
 
-export const Portals = () => {
+export const Portals = memo(() => {
```
Azein: So, it doesn't have any dependencies ever to re-consider the memo?

src/domains/chart/utils/formatters.ts
```diff
@@ -126,7 +129,7 @@ export const useFormatters = ({
   } = attributes
 
   const legendFormatValueDecimalsFromMinMax = useCallback((newMin: number, newMax: number) => {
-    if (newMin === min && newMax === max) {
+    if (safeEqualCheck(newMin, min) && safeEqualCheck(newMax, max)) {
```
Azein: Just a note, so it's not a left-to-right arguments order? Like (prevProps, nextProps), but the other way around?

src/domains/global/reducer.ts
```diff
@@ -219,6 +221,29 @@ globalReducer.on(setGlobalChartUnderlayAction, (state, { after, before, masterID
   },
 }))
 
+globalReducer.on(centerAroundHighlightAction, (state) => {
+  if (!state.globalChartUnderlay) {
```
Azein: Maybe it shouldn't dispatch an action if there is no global underlay. I don't have a strong opinion, however I'd try to avoid branching in reducer if it's possible.

src/domains/chart/components/portals.tsx
jacekkolasa: I don't understand, do you mean that it doesn't use any Props? This component will not have to update based on parent state.

src/domains/chart/utils/formatters.ts
jacekkolasa: oh yeah, that's a good idea to reverse it, haven't though of it. fixed

src/domains/global/reducer.ts
jacekkolasa: So this is just a defensive measure for the future to prevent potential bug, and also a way to make typing clear - because of the type:
```
state.globalChartUnderlay: null | {
 after: number
 ...
```
The action is not dispatched when there is no globalChartUnderlay.

Does it change your opinion?

PR #11, opened 2020-01-15
dygraph-fixes

jacekkolasa:
- user-friendly panning into the future,
- fixed edge case when switching to global-pan-mode during live-mode api-call,
- fixed reset on double-click and double-tap.
- updated jquery & bootstrap (security reasons)

### DIFF
--- a/src/domains/chart/components/abstract-chart.tsx
+++ b/src/domains/chart/components/abstract-chart.tsx
@@ -3,6 +3,7 @@ import classNames from "classnames"
 
 import { useDispatch } from "store/redux-separate-context"
 import { setGlobalChartUnderlayAction, setGlobalPanAndZoomAction } from "domains/global/actions"
+import { TimeRange } from "types/common"
 
 import { Attributes } from "../utils/transformDataAttributes"
 import {
@@ -45,7 +46,7 @@ interface Props {
   unitsCurrent: string
   viewAfter: number,
   viewBefore: number,
-  requestedViewRange: [number, number]
+  requestedViewRange: TimeRange
 }
 
 export const AbstractChart = ({
--- a/src/domains/chart/components/abstract-chart.tsx
+++ b/src/domains/chart/components/abstract-chart.tsx
@@ -45,6 +45,7 @@ interface Props {
   unitsCurrent: string
   viewAfter: number,
   viewBefore: number,
+  requestedViewRange: [number, number]
 }
 
 export const AbstractChart = ({
@@ -70,6 +71,7 @@ export const AbstractChart = ({
   unitsCurrent,
   viewAfter,
   viewBefore,
+  requestedViewRange,
 }: Props) => {
   const dispatch = useDispatch()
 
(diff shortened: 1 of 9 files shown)


src/domains/chart/components/abstract-chart.tsx
```diff
@@ -45,6 +45,7 @@ interface Props {
   unitsCurrent: string
   viewAfter: number,
   viewBefore: number,
+  requestedViewRange: [number, number]
```
builat: Could we use named tuples types? For this case it is totaly ok to write this way, but sometimes it is hard to understand
why not 
```typescript
number[]
but
[number, number]
```
Also as far as it is range, type of range could be used over the project.

src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -331,6 +316,44 @@ export const DygraphChart = ({
...
+    propsRef.current.resetGlobalPanAndZoom = resetGlobalPanAndZoom
+    propsRef.current.setGlobalChartUnderlay = setGlobalChartUnderlay
+    propsRef.current.viewAfter = viewAfter
+    propsRef.current.viewBefore = viewBefore
+  }, [chartData, globalChartUnderlay, hoveredX, resetGlobalPanAndZoom, setGlobalChartUnderlay,
+    viewAfter, viewBefore])
+
   useLayoutEffect(() => {
     if (chartElement && chartElement.current && !dygraphInstance) {
       const dygraphOptionsStatic = getInitialDygraphOptions({
```
builat: May be we can use this here?
`const {viewAfter, viewBefore} = propsRef.current`

src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -394,8 +417,9 @@ export const DygraphChart = ({
             const after = Math.round(xRange[0])
             const before = Math.round(xRange[1])
 
-            if (before <= (chartData.last_entry * 1000)
-              && after >= (chartData.first_entry * 1000)
+            const { first_entry: firstEntry, last_entry: lastEntry } = propsRef.current.chartData
+            if (before <= (lastEntry * 1000)
+              && after >= (firstEntry * 1000)
```
builat: could we use predicate named function for this if?

src/domains/chart/components/abstract-chart.tsx
jacekkolasa: good point! I'll change it to reuse a common type. Btw. i used it as tuple, just to have stricter type where it's possible

src/domains/chart/components/lib-charts/dygraph-chart.tsx
jacekkolasa: I don't understand, here i'm assigning necessary props to the `ref` because they are needed in old jquery library callbacks. I think i'm using destructuring few times here. Correct me please if i'm not on the same page :)

src/domains/chart/components/lib-charts/dygraph-chart.tsx
jacekkolasa: good idea, done https://github.com/netdata/dashboard/pull/11/commits/57e48c74ba6d5eed733f0a37a6616f310cb53160

PR #13, opened 2020-01-19
7729 print-view

jacekkolasa:
Print view rewroten
Based on the mode=print in the url, dashboard is fetching metrics for all charts, showing the progress of downloads and opening automatically native print screen


src/domains/chart/sagas.ts
```diff
@@ -3,67 +3,93 @@ import {
...
+    // to absolute global-pan-and-zoom, cancel the store update
+    // todo do xss check of data
+    if (action.type === fetchDataAction.success.toString()) {
+      const { viewRange } = action.payload.fetchDataParams
+      const globalPanAndZoom = (yield select(
+        selectGlobalPanAndZoom,
+      )) as GlobalStateT["globalPanAndZoom"]
+
+      if (globalPanAndZoom
+        && (viewRange[0] <= 0 || viewRange[1] <= 0)
```
builat: ma be could be destructured to for example `const [start, end] = viewRange`

src/domains/chart/sagas.ts
```diff
@@ -3,67 +3,93 @@ import {
...
+        continue
+      }
+    }
+
+    yield put(action)
+  }
+}
+
+
 type FetchDataSaga = { payload: FetchDataPayload }
```
builat: I think we could use common way to declare actions, like

```type MyAction<T = any> = { action: string, payload: T}```

In this case you will declare only generic in arguments. Here will be

```
function fetchDataSaga({ payload }: Action<FetchDataPayload>) {
```

File: src/domains/dashboard/utils/parse-url.ts [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+  alarm: null,
+  alarm_unique_id: 0,
+  alarm_id: 0,
+  alarm_event_id: 0,
+  alarm_when: 0,
+} as {[key: string]: unknown}
+
+const parseQueryPair = (queryTuple: [string, string]): {[key: string] : unknown} => {
+  const [key, value] = queryTuple
+  if (defaultUrlOptions[key] === undefined || value === undefined) {
```
builat: could be function

src/domains/dashboard/components/modal-portal.tsx
```diff
@@ -0,0 +1,20 @@
+import { ReactNode, useEffect, useRef } from "react"
+import { createPortal } from "react-dom"
+
+const modalRoot = document.getElementById("modal-root") as HTMLElement
```
VLegakis: @Azein @builat perhaps we could use a similar approach for our modal dialogs as @jacekkolasa does here

src/domains/chart/sagas.ts
jacekkolasa: good idea, i've reused "redux-act" `Action` type

PR #15, opened 2020-01-29
7767 settings migration

jacekkolasa:
https://github.com/netdata/netdata/issues/7767

dependent on https://github.com/netdata/dashboard/pull/14


src/main.js
```diff
@@ -3438,7 +3430,7 @@ function loadSnapshotDragAndDropSetup() {
     }
 };
 
-function loadSnapshotPreflight() {
+window.loadSnapshotPreflight = () => {
     var files = document.getElementById('loadSnapshotSelectFiles').files;
     if (files.length <= 0) {
```
builat: could ```files.length``` ever be less then 0? 
May be we can replace with if(!file.length)?

jacekkolasa: yes, definitely :) it was the old code

PR #16, opened 2020-02-05
misc-fixes

jacekkolasa:
various fixes for main.js. Because of a bug i needed to switch to load snapshots via separate sagas, i fixed some dependencies for Alarms modal, and done few other small tweaks

### DIFF
--- a/src/domains/chart/components/snapshot-loader.tsx
+++ b/src/domains/chart/components/snapshot-loader.tsx
@@ -5,6 +5,7 @@ import { serverDefault } from "utils/server-detection"
 import { selectIsSnapshotMode, selectSnapshotOptions } from "domains/dashboard/selectors"
 import { selectGlobalPanAndZoom } from "domains/global/selectors"
 import { useDispatch, useSelector } from "store/redux-separate-context"
+import { TimeRangeObjT } from "types/common"
 
 import { Attributes } from "../utils/transformDataAttributes"
 import { fetchDataForSnapshotAction } from "../actions"
@@ -26,8 +27,8 @@ const SnapshotLoader = ({
   const chartSettings = chartLibrariesSettings[chartLibrary]
 
   const globalPanAndZoom = useSelector(selectGlobalPanAndZoom)
-  const after = (globalPanAndZoom as { after: number, before: number }).after / MS_IN_SECOND
-  const before = (globalPanAndZoom as { after: number, before: number }).before / MS_IN_SECOND
+  const after = (globalPanAndZoom as TimeRangeObjT).after / MS_IN_SECOND
+  const before = (globalPanAndZoom as TimeRangeObjT).before / MS_IN_SECOND
 
   const dispatch = useDispatch()
   useEffect(() => {
--- a/src/domains/chart/components/snapshot-loader.tsx
+++ b/src/domains/chart/components/snapshot-loader.tsx
@@ -1,8 +1,15 @@
-import React from "react"
+import React, { useEffect } from "react"
 
-import { Attributes } from "domains/chart/utils/transformDataAttributes"
-import {useSelector} from "store/redux-separate-context"
-import {selectIsSnapshotMode} from "domains/dashboard/selectors"
+import { MS_IN_SECOND } from "utils"
+import { serverDefault } from "utils/server-detection"
+import { selectIsSnapshotMode, selectSnapshotOptions } from "domains/dashboard/selectors"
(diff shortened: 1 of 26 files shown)


src/domains/chart/selectors.ts
```diff
@@ -61,3 +61,19 @@ export const selectNameOfAnyFetchingChart = createSelector(
   (chartsState) => Object.values(chartsState)
     .find((chartState) => chartState.isFetchingData)?.chartDetails?.id,
 )
+
+export const selectAmountOfSnapshotsFetched = createSelector(
+  selectChartsState,
+  (chartsState) => Object.values(chartsState)
+  // count the nr of "success" or "failure" charts
+    .map((chartState) => Boolean(chartState.snapshotData))
+    .reduce((acc, value) => acc + Number(value), 0),
```
Azein: I'm not sure I understand. So, the `value` is a boolean type (from prev `map`), and we cast it to Number?

src/domains/chart/utils/get-chart-url-options.ts
```diff
@@ -0,0 +1,28 @@
+import { Attributes } from "domains/chart/utils/transformDataAttributes"
+import { chartLibrariesSettings } from "domains/chart/utils/chartLibrariesSettings"
+
+export const getChartURLOptions = (
```
Azein: The string concat into a variable is a valid technique, maybe not worth to redo this function to be fancier/using templates with function calls to get specific string parts. Still pointed out, consider as a nitpick.

src/domains/chart/components/snapshot-loader.tsx
```diff
@@ -0,0 +1,70 @@
...
+}: SnapshotLoaderProps) => {
+  const host = attributes.host || serverDefault
+  const { snapshotDataPoints } = useSelector(selectSnapshotOptions)
+  const group = attributes.method || window.NETDATA.chartDefaults.method
+  const { chartLibrary } = attributes
+  const chartSettings = chartLibrariesSettings[chartLibrary]
+
+  const globalPanAndZoom = useSelector(selectGlobalPanAndZoom)
+  const after = (globalPanAndZoom as { after: number, before: number }).after / MS_IN_SECOND
+  const before = (globalPanAndZoom as { after: number, before: number }).before / MS_IN_SECOND
```
builat: `{ after: number, before: number }`

could be `type RangeT = { after: number, before: number }`

src/domains/chart/selectors.ts
jacekkolasa: yes, i need the number of `chartStates` where the boolean is true, do you think it could be done differently, so it's more clear?
> // count the nr of "success" or "failure" charts

But this ^^ comment is wrong, it's a result of copy/paste, that's probably what is confusing

src/domains/chart/utils/get-chart-url-options.ts
jacekkolasa: i agree! i left it like this because how it was initially working in old dashboard.js, definitely something which can be shortened and more readable

src/utils/netdata-sdk/metrics-stream.ts
```diff
@@ -1,36 +1,43 @@
...
+  mergeMap, tap, catchError, startWith, switchMap,
 } from "rxjs/operators"
 
 import { axiosInstance } from "./axios-instance"
 
-const CONCURRENT_CALLS_LIMIT = 20
-
 interface FetchInputEvent {
   url: string
   params: { [key: string]: unknown }
```
builat: could be `type UnknownStringKeyT =  { [key: string]: unknown }`

src/domains/chart/selectors.ts
builat: so why not to do than in one `reduce` iteration?

builat: `.reduce((acc, chartState) => acc + !!chartState ? 1 : 0 , 0),`

src/utils/netdata-sdk/metrics-stream.ts
jacekkolasa: i like that idea. fixed

src/domains/chart/components/snapshot-loader.tsx
jacekkolasa: done

src/domains/chart/selectors.ts
jacekkolasa: yes, that looks better! fixed

PR #17, opened 2020-02-07
260 navigation bar

Azein:
Contains application header. Some supposedly dynamic data is held in `mock.ts`. Brings in `styled-components`, `@netdata/netdata-ui`, `craco` webpack override for svg sprites, and I also added `prettierrc`. Code itself is more about layout, I used some old icons for export/import, stuff should be responsive up to 400-something px.


src/App.tsx
```diff
@@ -1,98 +1,102 @@
-import React, {
-  useEffect, useLayoutEffect, useRef, useState,
-} from "react"
-import Ps from "perfect-scrollbar"
+import React, { useEffect, useLayoutEffect, useRef, useState } from "react";
```
jacekkolasa: can you revert the "semis" changes?

jacekkolasa:
thanks for bringing that additional stuff and configuration. the navbar looks neat, it's nice to see it live :)

One thing, i'll point it to different branch, because we will first need to release Dashboard with old navbars

src/App.tsx
Azein: Sure, I think it was the change before I noticed the prettierrc problem

PR #18, opened 2020-02-08
Space sidebar

Azein:
Provides visuals for spaces bar (with initials), and space panel. Panel is managed with redux state, and utility saga also adds/removes some classname from `body` to keep layout consistent - with some breakpoints.
For data I went with rough mocks, and pasted some commented types from cloud project as reference if needed on further stages of development.
One strange problem I had, was that something is setting `text-shadow` on the `h` elements. Used override, but why it happens I don't know.

### DIFF
--- a/src/components/space-panel/components/space-rooms/room-label/room-label.tsx
+++ b/src/components/space-panel/components/space-rooms/room-label/room-label.tsx
@@ -1,0 +1,49 @@
+import React from "react"
+import {
+  StyledIcon,
+  Container,
+  RoomName,
+  IndicatorsContainer,
+  ErrorIndicator,
+  WarningIndicator,
+  UnreachableIndicator,
+} from "./styled"
+
+/* Counters are commented out, but might be used, UI is ready - if we have data */
+// const getAlarmsCount = (count: number) => (count > 99 ? "99+" : `${count}`)
+
+interface Props {
+  room: any
+}
+
+export const RoomLabel = ({ room }: Props) => {
+  const alarmCounter = room.alarmCounter || ({} as any)
+
+  const handleSelectRoom = () => {
+    // navigate to cloud room
+  }
+
+  return (
+    <Container onClick={handleSelectRoom}>
+      <StyledIcon name="room" />
+      <RoomName>{room.name}</RoomName>
+      <IndicatorsContainer>
+        {alarmCounter?.critical > 0 && (
+          <ErrorIndicator>
+            {/* <StyledCount>{getAlarmsCount(alarmCounter.critical)}</StyledCount> */}
+          </ErrorIndicator>
+        )}
+        {alarmCounter?.warning > 0 && (
+          <WarningIndicator>
+            {/* <StyledCount>{getAlarmsCount(alarmCounter.warning)}</StyledCount> */}
+          </WarningIndicator>
+        )}
+        {alarmCounter?.unreachable > 0 && (
+          <UnreachableIndicator>
+            {/* <UnreachableCount>{getAlarmsCount(alarmCounter.warning)}</UnreachableCount> */}
+          </UnreachableIndicator>
+        )}
+      </IndicatorsContainer>
+    </Container>
+  )
+}
--- a/src/components/space-panel/components/visited-nodes/visited-nodes.tsx
+++ b/src/components/space-panel/components/visited-nodes/visited-nodes.tsx
@@ -1,10 +1,16 @@
 import React from "react"
 import { CollapsibleList, SimpleListItem } from "@rmwc/list"
-import { Text } from "@netdata/netdata-ui"
 import "@material/list/dist/mdc.list.css"
 import "@rmwc/list/collapsible-list.css"
 import "@rmwc/icon/icon.css"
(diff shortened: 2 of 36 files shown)


src/components/space-panel/components/visited-nodes/visited-nodes.tsx
```diff
@@ -0,0 +1,67 @@
...
+    }
+  }
+
+  return (
+    <NodesContainer>
+      <CollapsibleList
+        startOpen
+        handle={
+          <ListHeaderContainer>
+            <SimpleListItem metaIcon="chevron_right" text="Visited Nodes" />
```
jacekkolasa: i see a strange issue that clicking this component (or other nodes in the sidebar) is shifting the right-side sidebar upwards/downwards. Do you know something about it? Perhaps it's a old dashboard issue, i'm not sure (if it is, i can look at it later)

src/components/space-panel/components/visited-nodes/visited-nodes.tsx
```diff
@@ -0,0 +1,67 @@
...
+      window.open(`http://${url}`, "_blank")
+    }
+  }
+
+  return (
+    <NodesContainer>
+      <CollapsibleList
+        startOpen
+        handle={
+          <ListHeaderContainer>
```
jacekkolasa: on hover the background has no horizontal padding, shouldn't it cover the whole width (at least on the left)?
image

src/components/space-panel/components/space-rooms/room-label/room-label.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const alarmCounter = room.alarmCounter || ({} as any)
+
+  const handleSelectRoom = () => {
+    // navigate to cloud room
+  }
+
+  return (
+    <Container onClick={handleSelectRoom}>
+      <StyledIcon name="room" />
+      <RoomName>{room.name}</RoomName>
```
jacekkolasa: shouldn't they also have the same :hover effect as visited nodes?

src/index.css
```diff
@@ -1,13 +1,24 @@
...
+}
+
+@media (min-width: 325px) {
+  .with-panel {
+    padding-left: 56px;
+  }
+}
+@media (min-width: 1200px) {
+  .with-panel {
+    padding-left: 280px;
```
jacekkolasa: The effect is nice! :) and it's good that for smaller screens it doesn't change the body width. But you added `padding-left: 56px` both for "body" and small .with-panel

jacekkolasa:
it feels good to see Dashboard changing its look after few years, nice work! Most comments are details, but probably we should fix the hover and make it consistent

src/components/space-panel/components/visited-nodes/visited-nodes.tsx
Azein: If you click anywhere else, it will still happen. So it's connected to sidebar.

src/components/space-panel/components/visited-nodes/visited-nodes.tsx
Azein: Maybe it should, however horizontal padding is also used for scrollbar.  This hover comes with rmwc so I have no guidelines for that.

src/components/space-panel/components/space-rooms/room-label/room-label.tsx
Azein: I don't see hovers at all in our designs. Ones that have it - it's because of rmwc defaults. I won't override anything unless there is a direct product issue with this behaviour.

src/index.css
Azein: 56px is for workspaces sidebar, I guess it should persist even on smaller screens.

src/components/space-panel/components/space-rooms/room-label/room-label.tsx
jacekkolasa: We should probably ask Javier about it. I commented on this issue because you have more experience with those components, and i'm not sure if we won't have to change how paddings work here (no padding on container, paddings on individual items)

src/index.css
jacekkolasa: what i mean is
```
@media (min-width: 325px) {
  .with-panel {
    padding-left: 56px;
  }
}
```
is a duplicate of 
```
body {
  padding-left: 56px;
}
``` ;)

File: src/components/space-panel/components/visited-nodes/visited-nodes.tsx [hunk 5]
```diff
@@ -0,0 +1,67 @@
...
+      }
+    >
+      {urls.map((url: string, i: number) => (
+        <ListItem
+          key={i}
+          onClick={() => {
+            visitNode(url)
+          }}
+        >
+          <NodeUrl>{url}</NodeUrl>
```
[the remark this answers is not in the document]
Azein: This is dropped according to design.

src/components/space-panel/components/visited-nodes/visited-nodes.tsx
Azein: Fixed.

PR #19, opened 2020-02-12
Node View Component

jacekkolasa:
Node View Component introduced - right now to be used in Cloud.


src/domains/dashboard/components/head-main.tsx
```diff
@@ -0,0 +1,266 @@
...
+export const HeadMain = ({
+  charts,
+  duration,
+}: Props) => {
+  const commonAttributes = {
+    after: -duration,
+    points: duration,
+  }
+
+  // todo
```
Azein: So, what's the plan? We don't have this object in cloud, obviously =)

src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -0,0 +1,307 @@
...
+  chartsMetadata: ChartsMetadata
+}
+export const NodeView = ({
+  chartsMetadata,
+}: Props) => {
+  const [width, setWidth] = useState(0)
+  const ref = useRef<HTMLDivElement>(null)
+  useEffect(() => {
+    if (!width && ref.current) {
+      setWidth(ref.current.getBoundingClientRect().width)
```
Azein: So, it sets correctly with single `width` as effect dependency? Not callback ref needed? I always struggle with these things.

src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -0,0 +1,307 @@
...
+    ((((width * pcentWidth) / 100) * chartsMetadata.update_every) / 3) / 60,
+  ) * 60
+
+  const menus = renderChartsAndMenu(chartsMetadata)
+  const main = sortObjectByPriority(menus)
+
+  return (
+    <div className="node-view__container">
+      <div ref={ref} className="charts-body" role="main">
+        {!!duration && (
```
Azein: duration is a number, right? Does `!!` make sense?

src/domains/dashboard/components/head-main.tsx
jacekkolasa: well, you have this, please check again for `window.NETDATA.themes` :)
i've put it before, when i was adding charts (and you had it previously, too, when dashboard.js was there). It's of course `todo`, but it's working and it's making few things easier with static custom dashboards

https://github.com/netdata/dashboard/blob/f18b152b48ea18d687b3e3ddc2ebce28e52e718b/src/index-npm.ts#L10

src/domains/dashboard/components/node-view/node-view.tsx
jacekkolasa: are you asking if it's ok to not provide `setWidth` as a effect dependency? as far as i remember the second item from `useEffect()` array is always the same (like useRef).

And i set this width only once on purpose (because that's how old Dashboard was working, we can add features later)

src/domains/dashboard/components/node-view/node-view.tsx
jacekkolasa: yes, IMO for number it's a must have, didn't you never see bugs with rendered `0`s in the past? :)

src/domains/chart/components/chart-container.tsx
```diff
@@ -10,23 +10,23 @@ export type Props = {
   attributes: Attributes
   // warning! this is not the same as chartId in old dashboard
   // here, the chartID must be unique across all agents
-  chartID: string
+  chartUuid: string
```
VLegakis: Minor note we use the ID convention nomatter if its a uuid or other. Not important. just noting...

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -75,11 +75,11 @@ type AttributeConfig = BaseAttributeConfig | BooleanAttributeConfig
 
 export interface Attributes {
   id: string
-  host: string | undefined
+  host?: string | undefined
```
VLegakis: If its optional I think the undefined type is not needed. Minor thing...

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -75,11 +75,11 @@ type AttributeConfig = BaseAttributeConfig | BooleanAttributeConfig
...
   title?: string
   chartLibrary: ChartLibraryName
   width: number | string | null
-  height: number | string | null
+  height?: number | string | null
```
VLegakis: Same here for optionals undefined is assumed (not sure if you need it as null here though...)

src/domains/chart/utils/transformDataAttributes.ts
jacekkolasa: good catch, that shouldn't accept `undefined` anymore, thx

src/domains/chart/components/chart-container.tsx
jacekkolasa: so the problem here is that for the Agent, chartID (chart.id) is a metric, and we could actually have multiple charts rendered for the same metric - for example Dygraph and Gauge charts for "system.cpu". So to differentiate it i thought we could have different naming.

@VLegakis what do you think? This will be also a thing to remember when putting charts into Charts Table

src/domains/chart/utils/transformDataAttributes.ts
jacekkolasa: @VLegakis `null` can actually be supported in some cases (in static custom dashboards) and have different behaviour (backwards compatibility), but thanks for asking! i will try narrow down those possibilities in the future, but i suspect that `null` in old static-custom-dashboards was resetting the height after resizing

PR #20, opened 2020-02-14
node-view performance fix + misc

jacekkolasa:
- fixed node view performance (separate scrolling/active-chart state from charts)
- hooked scroll position to active menu
- fixed few tests
- bumped package version
- how to use NodeView (small example) - https://github.com/netdata/dashboard/blob/node-view-hash-hook/src/domains/dashboard/components/node-view/README.md


src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -208,14 +216,31 @@ export const NodeView = ({
   }, [width])
 
 
+  const { y } = useWindowScroll()
+  useEffect(() => {
+    if (ref.current) {
+      const currentNode = Array.from(ref.current.querySelectorAll("[id]"))
+        .filter((node) => (node.getAttribute("id") as string).startsWith("menu"))
+        .find((node) => node.getBoundingClientRect().top > 0)
```
builat: filter could be not used here due to correct function in find. 
For example
```
const menuWithRect => (node: any) =>  {
(node.getAttribute("id") as string).startsWith("menu")) && 
node.getBoundingClientRect().top > 0
}
...........
Array.from(ref.current.querySelectorAll("[id]")).find(menuWithRect)
```

src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -104,21 +107,22 @@ const SubSection = ({
             )),
         )}
       </div>
-      {submenuNames.map((submenu: string) => {
-        const submenuID = name2id(`menu_${menu}_submenu_${submenu}`)
-        const chartsSorted = menu.submenus[submenu].charts
+      {submenuNames.map((submenuName: string) => {
```
builat: Maybe we could separate this map function into different file?

src/domains/dashboard/components/menu-sidebar.tsx
```diff
@@ -0,0 +1,90 @@
...
+                <span dangerouslySetInnerHTML={{ __html: menu.icon }} />
+                {" "}
+                {menu.title}
+              </a>
+              <ul
+                className={classNames(
+                  "node-view__sidebar-submenu",
+                )}
+              >
+                {submenuNames.map((submenuName) => {
```
builat: May be we can split this map function into different file?

src/domains/dashboard/components/menu-sidebar.tsx
```diff
@@ -0,0 +1,90 @@
...
+      const node = document.getElementById(chartNodeId.replace("#", ""))
+      if (node) {
+        node.scrollIntoView()
+      }
+    }
+  }
+
+
+  return (
+    <div className="node-view__sidebar" role="complementary">
```
VLegakis: Nice! Just for curiosity what is this `role` used here?

jacekkolasa: @VLegakis good tip about the redux-store! I changed it to "Dashboard Charts".
The `role`s are copied from the old Dashboard, but i had to change some of them (eslint was complaining). I'm still not an expert on them, but i learned something.  Those attributes are used for screen readers and magnifiers

src/domains/dashboard/components/node-view/node-view.tsx
jacekkolasa: Sure, i can do that, do you mean function or separate component? But btw. don't you think it will a little complicate this, because you will have to open 2 files to actually read that and create map in head to comprehend it? The function will never be reused btw.

src/domains/dashboard/components/node-view/node-view.tsx
jacekkolasa: good idea to separate that function. Not sure if merging is better than composition of 2, but it's not that complicated, so all good

src/domains/dashboard/components/menu-sidebar.tsx
VLegakis: Gooood. Didn't know that! thanks!

Issue #26, opened 2020-02-26
create a release on new git tag


jacekkolasa:
When new git tag is created, CI/CD should trigger a pipeline:

`npm install` (it's important that it's npm, not yarn)
`npm run build`

The contents of the built `/build` folder should be part of the release.
Checksum file should be built, to validate the compilation

PR #27, opened 2020-02-26
Add basic release build automation.

Ferroin:
When a new release is created, this will automatically build the dashboard and add a rlease asset containing a compressed tarball of the build directory.

Long-term, this can be easily expanded to publish to NPM (if we choose to do so) and open a PR against https://github.com/netdata/netdata to update the packaging code there to use the newly released version of the dashboard.

Closes: #26

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,31 @@
+---
+name: Release
+on:
+  release:
+    type: [created]
+jobs:
+  build:
+    name: 'Build and Publish Artifacts'
+    runs-on: ubuntu-latest
+    stages:
+      - name: 'Checkout Sources'
+        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Ferroin:
@jacekkolasa I'm operating under the assumption that once this is merged we'll cut a new release (well, pre-release, but you get the idea) so that we have one with the updated artifacts that we can work against for the other stuff that's needed to get this working with the agent.

jacekkolasa:
Yes, that will work. I'll have few bugfixes here EOD tomorrow, but the current version on `master` branch should work when uploaded to web/ folder in Netdata

prologic:
@Ferroin Hmmm 🤔Thanks for figuring this out; I'll change the way we do releases for eBPF to match this; it is better to split this out in cases of transient failures uploading release assets where you can just re-run _that_ workflow.

PR #31, opened 2020-03-02
Pre release fixes pack

jacekkolasa:
a pack of fixes for first dashboard release (beta). when reviewing, i think it's better to do that commit-by-commit, because otherwise the changes are not related to each other

### DIFF
--- a/src/hooks/use-show-value-outside.ts
+++ b/src/hooks/use-show-value-outside.ts
@@ -1,5 +1,6 @@
 import { useEffect, useRef } from "react"
 import { isEmpty } from "ramda"
+import { useMount } from "react-use"
 
 import { ChartData, DygraphData } from "domains/chart/chart-types"
 import { Attributes } from "domains/chart/utils/transformDataAttributes"
@@ -30,7 +31,7 @@ export const useShowValueOutside = ({
   const showValueAttributesNodes = useRef<(HTMLElement | null)[]>([])
 
   // find the nodes that will have populated values
-  useEffect(() => {
+  useMount(() => {
     const { showValueOf } = attributes
     // showValueOf will be undefined if not used, but additional isEmpty check can prevent
     // regression performance issue in the future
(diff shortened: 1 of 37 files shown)


src/hooks/use-show-value-outside.ts
```diff
@@ -0,0 +1,80 @@
...
+    const { showValueOf } = attributes
+    // showValueOf will be undefined if not used, but additional isEmpty check can prevent
+    // regression performance issue in the future
+    if (!showValueOf || isEmpty(showValueOf)) {
+      return
+    }
+    const dimensionNames = chartData.dimension_names
+    const dimensionIds = chartData.dimension_ids
+    dimensionNames.forEach((dimensionName, i) => {
+      const userElementId = showValueOf[`show-value-of-${dimensionName.toLowerCase()}`]
```
Azein: Could be an external util, maybe this level of detail how we call the key isn't so important.

src/hooks/use-show-value-outside.ts
```diff
@@ -0,0 +1,80 @@
...
+        || showValueOf[`show-value-of-${dimensionIds[i].toLowerCase()}-at`]
+
+      // if element is not found, just add null
+      showValueAttributesNodes.current = showValueAttributesNodes.current.concat(
+        document.getElementById(userElementId),
+      )
+    })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []) // do only once on purpose
```
Azein: I think, react-use has an effect for this, which doesn't require eslint disable / additional commenting. Like useDidMount or something.

src/hooks/use-show-value-outside.ts
jacekkolasa: do you mean that `show-value-of-${dimension}-at` is too descriptive? This is the exact name of an attribute from custom dashboards

src/hooks/use-show-value-outside.ts
jacekkolasa: good idea, will do that now

src/hooks/use-show-value-outside.ts
Azein: Well, technically of course this is needed, I'm saying that probably it's not required to see the exact string templates when reading the code there.

Issue #32, opened 2020-03-02
Issues detected on dashboard


thiagoftsm:
I executed the following tests using Google Chrome and Mozilla Firefox:

- I moved from top to bottom Netdata dashboard to verify that all charts were displayed correctly.
- I used the direct zoom on charts using keyboard and I also used `+` and `-` to create zooms
- I tried to open alarms to visualize them.
- I raised alarms to verify that alarm centralization was working.
- I had one slave running to  be sure that the move from one host to other would not create problems with the charts.
- I hide dimensions and select multiple dimensions on different charts
- I repeat the previous tests with browser console opened to check possible problems with JS variables.

During the tests with latest dashboard the following issues were found:

- [x] Small charts used to describe cpu utilization are on top of text.
- [x] When I try to open Alarms using dashboard, the first time that I click on link it raises the error `Cannot load the required JS library: bootstrap-table`, but when I close the message and the alarm and I try to access again, everything works.
- [x] On the right side menu, `eBPF` and `Applications` has the same image, but `eBPF` is not showing.
- [x] When dashboard tries to access `badge.svg?chart=system.cpu&alarm=10min_cpu_iowait&refresh=auto:1` I receive `NS_ERROR_FAILURE`.
- [x] I am detecting in different times the message `scrollToId is not defined` on events onclick. I received now a different message on this variable`Uncaught ReferenceError: scrollToId is not defined`

thiagoftsm:
Errors detected on 0.2.4:

- [ ] Due the fact we had a bug when I tried to open the alarm, I could not test the alarm on dashboard properly, now that it is working fine, I detected that when I click on alarm log, Netdata is not redirecting to another tab and of course I do not have the alarm centralized on screen.
- [ ] I am seeing these errors now : 
```
   Uncaught TypeError: Cannot read property 'isUsingGlobalRegistry' of undefined
    at Object.onClickRow (:19999/static/js/8.a36d9b89.chunk.js:2)
    at d.trigger (:19999/static/js/13.b3f47c79.chunk.js:1)
    at HTMLTableCellElement.<anonymous> (:19999/static/js/13.b3f47c79.chunk.js:1)
    at HTMLTableCellElement.dispatch (:19999/static/js/0.2dfc59a1.chunk.js:2)
    at HTMLTableCellElement.v.handle (:19999/static/js/0.2dfc59a1.chunk.js:2)
`` 
- [ ] This was only a warning, please, take a look whether it is necessary to change something:
```
[...]

PR #33, opened 2020-03-04
sign-in button and chart height fixes

jacekkolasa:
- fixes sign-in button when using private registry - https://github.com/netdata/dashboard/issues/32#issuecomment-594098230
- fixes initial chart height when using localStorage data from old dashboard

### DIFF
--- a/src/domains/chart/utils/get-portal-node-styles.ts
+++ b/src/domains/chart/utils/get-portal-node-styles.ts
@@ -37,7 +37,10 @@ export const getPortalNodeStyles: GetPortalNodeStyles = (
     : null
 
   if (heightFromLocalStorage) {
-    height = heightFromLocalStorage
+    // .replace() is for backwards compatibility -  old dashboard was always doing
+    // JSON.stringify when setting localStorage so many users have '"180px"' values set.
+    // We can remove .replace() after some time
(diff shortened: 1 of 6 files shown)


src/domains/chart/utils/get-portal-node-styles.ts
```diff
@@ -37,7 +37,10 @@ export const getPortalNodeStyles: GetPortalNodeStyles = (
     : null
 
   if (heightFromLocalStorage) {
-    height = heightFromLocalStorage
+    // .replace() is for backwards compatibility -  old dashboard was always doing
+    // JSON.stringify when setting localStorage so many users have '"180px"' values set.
+    // We can remove .replace() after some time
+    height = heightFromLocalStorage.replace(/"/g, "")
```
builat: Could't be there something like `"\"some\""`?

jacekkolasa: @builat no, it's just for this specific property type. if you go to any demo netdata page and do `localStorage["chart_heights.frankfurt_netdata_rocks_system_cpu"]`, you'll get `""180px""`

jacekkolasa: it's just for reading properties saved by old dashboard. new dashboard will just save normal strings, without quotes

PR #36, opened 2020-03-15
8363-perf-fixes

jacekkolasa:
- some performance fixes, regarding scrolling in Dashboard/Node-View. Generally i've put more effort to not update dygraph instance too often after mount. Unfortunately it's at the cost of putting more stuff into refs :/
- exposed global-pan-and-zoom action, to be used by Node View (we'll talk about it)
- small bugfix in data-show-value-outside hook

### DIFF
--- a/src/domains/chart/components/lib-charts/dygraph-chart.tsx
+++ b/src/domains/chart/components/lib-charts/dygraph-chart.tsx
@@ -1,7 +1,8 @@
 import { sortBy } from "ramda"
 import React, {
-  useLayoutEffect, useRef, useCallback,
+  useLayoutEffect, useRef, useCallback, RefObject,
 } from "react"
+import { useUpdateEffect } from "react-use"
 import Dygraph from "dygraphs"
 import "dygraphs/src-es5/extras/smooth-plotter"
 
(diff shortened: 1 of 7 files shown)


src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -456,6 +466,27 @@ export const DygraphChart = ({
...
 
+        underlayCallback(canvas: CanvasRenderingContext2D, area: DygraphArea, g: Dygraph) {
+          // the chart is about to be drawn
+          // this function renders global highlighted time-frame
+
+          if (propsRef.current.globalChartUnderlay) {
+            const { after, before } = propsRef.current.globalChartUnderlay
+
+            if (after < before) {
+              const bottomLeft = g.toDomCoords(after, -20)
```
builat: What is this `20`?

jacekkolasa: That was copy/pasted code from different place, but overall good point, i fixed it! 
 But i left the `const HIGHLIGHT_HORIZONTAL_PADDING = ` right above that code, so it's easy to red, if that's ok

PR #37, opened 2020-03-17
NodeView fixes

VLegakis:
Starting getting used to the code base

### DIFF
--- a/src/domains/chart/components/chart-with-loader.tsx
+++ b/src/domains/chart/components/chart-with-loader.tsx
@@ -74,7 +74,16 @@ export const ChartWithLoader = ({
         host,
       }))
     }
-  }, [attributes.id, chartUuid, dispatch, host, isFetchingDetails, chartMetadata, externalChartMetadata])
+  },
+  [
+    attributes.id,
+    chartUuid,
+    dispatch,
+    host,
+    isFetchingDetails,
+    chartMetadata,
+    externalChartMetadata,
+  ])
 
 
   // todo local state option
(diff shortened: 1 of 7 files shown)


src/domains/dashboard/components/head-main.tsx
```diff
@@ -25,6 +27,7 @@ export const HeadMain = ({
         <ChartWrapper
           attributes={{
             ...commonAttributes,
+            host,
```
jacekkolasa: can we please add `host` to commonAttributes?

VLegakis: Done

src/domains/chart/components/chart-with-loader.tsx
```diff
@@ -47,31 +41,29 @@ export type Props = {
...
    */
   const host = attributes.host || serverDefault
   const dispatch = useDispatch()
-  const selectChartMetadataRequest = useMemo(makeSelectChartMetadataRequest, [])
-  const { chartMetadata, isFetchingDetails } = useSelector((state: AppStateT) => (
-    selectChartMetadataRequest(state, { chartId: attributes.id, id: chartUuid })
-  ))
   useEffect(() => {
-    if (!chartMetadata && !isFetchingDetails) {
+    if (!chartMetadata) {
```
jacekkolasa: i think we misunderstood - this component is used also by the Agent Dashboard, so we cannot make any breaking changes here, we just need to handle scenario when `chartMetadata` is passed as a prop. We need to handle `redux` still (for Dashboard), so we'll need to have those lines back (`isFetchingDetails`, `selectChartMetadataRequest`, etc.)

PR #38, opened 2020-03-18
pre-release-fixes-2

jacekkolasa:
gauge & easypiechart width update
 proper centering when clicking alarm in alarms-log
 keep loader until react app is ready
 fix top bar not showing when highlighting (few lines of code)
 "check known URLs" issue
 "reset" options is not hooked up
 go to host node from streamed node (low priority, because that's removed on release)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -961,7 +961,7 @@ function gotoServerValidateUrl(id, guid, url) {
     setTimeout(function () {
         document.getElementById('gotoServerList').innerHTML += '<tr><td style="padding-left: 20px;"><a href="' + verifyURL(finalURL) + '" target="_blank">' + escapeUserInputHTML(url) + '</a></td><td style="padding-left: 30px;"><code id="' + guid + '-' + id + '-status">checking...</code></td></tr>';
 
-        NETDATA.registry.hello(url, function (data) {
+        NETDATA.registryHello(url, function (data) {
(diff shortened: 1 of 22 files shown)


src/main.js
```diff
@@ -972,7 +973,8 @@ function gotoServerValidateUrl(id, guid, url) {
                     if (gotoServerMiddleClick) {
                         window.open(verifyURL(finalURL), '_blank');
                         gotoServerMiddleClick = false;
-                        document.getElementById('gotoServerResponse').innerHTML = '<b>Opening new window to ' + NETDATA.registry.machines[guid].name + '<br/><a href="' + verifyURL(finalURL) + '">' + escapeUserInputHTML(url) + '</a></b><br/>(check your pop-up blocker if it fails)';
+                        const registryMachines = getFromRegistry("registryMachines");
+                        document.getElementById('gotoServerResponse').innerHTML = '<b>Opening new window to ' + registryMachines[guid].name + '<br/><a href="' + verifyURL(finalURL) + '">' + escapeUserInputHTML(url) + '</a></b><br/>(check your pop-up blocker if it fails)';
```
builat: Could this markdown be wrapped into template or function?

jacekkolasa: yeah, i agree completely, but we have lots of strings like this in main.js... we'll remove all of them someday when we'll clean (remove) the whole main.js file

builat: Oh, ok!

PR #54, opened 2020-03-30
private-registry-navbar-fixes

jacekkolasa:


dashboard-react.js
```diff
@@ -590,7 +590,82 @@ NETDATA.registryHello = function (host, callback) {
...
-    NETDATA.error(407, host);
+    // NETDATA.error(407, host);
+
+    if (typeof callback === 'function') {
+      return callback(null);
+    }
+  });
+}
+
+NETDATA.registrySearch = function (machine_guid, getFromRegistry, serverDefault, callback) {
```
jacekkolasa: those are copies from old registry code, we'll refactor this when we'll get rid of main.js

PR #55, opened 2020-03-30
Migrate to latest ACLK spec`ed behaviour

VLegakis:
Based from discussions in: https://github.com/netdata/product/issues/277

Still the Notifications UI doesn't match the specified one
https://www.figma.com/file/28Xv1J95zLbPowJujLuArZ/%F0%9F%93%A6Netdata-Design-System-0.7?node-id=755%3A0
If we have time we can bring this to par.


amoss:
How do I test this?

src/domains/chart/sagas.ts
```diff
@@ -241,38 +241,53 @@ function* fetchChartSaga({ payload }: Action<FetchChartPayload>) {
...
   const { poll } = payload
   let isCloudEnabled = false
-  let isCloudClaimed = false
+  let isAgentClaimed = false
+  let isCloudAvailable = false
+  let isACLKAvailable = false
 
   try {
     const registry = yield select(selectRegistry)
     const wasCloudAvailable = registry?.isCloudAvailable || false
```
amoss: I can't see where this gets changed. What about the case where this is false the first time it is read, but then it becomes true and then false again later?

jacekkolasa:
> How do I test this?

@amoss you can test in dev mode, by:
- ensuring you have `nodejs/npm` installed
- running `npm install && npm run start:dashboard`
- navigate to localhost:3000 (and you must have netdata under localhost:19999 on that machine)

or you can just test it when we'll release it to dashboard/master branch and make a PR in netdata repo.

amoss:
I can test it when the PR updates it in the main repo. Do I need to review code in the dashboard repo? I'm wondering if one of the front-end team would be a more appropriate reviewer for the typescript code.

jacekkolasa:
I agree, it's not needed, i think your review is just optional. I think review will be more needed in netdata repo, as an end user opinion :)

PR #57, opened 2020-04-01
fix-cloud-timeouts

jacekkolasa:
set timeout for metrics-data calls, to 15s
attributes.pointsPerPixes is now overwriting existing settings even if it's lower (previously it was Math.max() with settings/defaults)
allowed more simultaneus calls (60 for cloud)

### DIFF
--- a/src/utils/netdata-sdk/metrics-stream.ts
+++ b/src/utils/netdata-sdk/metrics-stream.ts
@@ -16,6 +16,8 @@ interface FetchInputEvent {
   onSuccessCallback: (data: { [key: string]: unknown }) => void
 }
 
+const METRICS_TIMEOUT = 15_000
+
 export const getFetchStream = (concurrentCallsLimit: number) => {
   const fetch$ = new Subject<FetchInputEvent>()
   const resetFetch$ = new Subject()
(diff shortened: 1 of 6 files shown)


src/utils/netdata-sdk/metrics-stream.ts
```diff
@@ -16,14 +16,16 @@ interface FetchInputEvent {
   onSuccessCallback: (data: { [key: string]: unknown }) => void
 }
 
+const METRICS_TIMEOUT = 15_000
```
VLegakis, line 19: Strange annotation :) I didn't know about it

jacekkolasa: yeah, it came quite recently into typescript :)

jacekkolasa:
> So, the timeout per poll is 15 secs?
yes, do you think it should be smaller? We talked about 10s recently, so we can lower it down, but as @lvrach said it should be product decision. So I just set a value which IMO is better than no-timeout :)

> In case it gets errors (upon first call or after that) will it stop polling?
No, actually the chart is still trying, so we're good. The error would have happened if we didn't had any metadata for chart, but we do, so Dashboard keeps polling.

> At the same frequency?
It retries every second, but if we want to make it less often, we can tweak that with attribute in the cloud.

> All the above are the same between the cloud and Agent variants of the Dashboard or it behaves exactly the same?
Yes. Just with exception of concurrent-calls-limit. That's because you can see much more charts at the same time (in table view)

PR #58, opened 2020-04-03
remove notifications and fix /default.html route

jacekkolasa:
@VLegakis i disabled the notifications for now, i know that your PR contains the fix, but i don't have time for testing this, let's just not have notifications and /api/v1/info polling at all next stable release (after weekend)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "0.4.10",
+  "version": "0.4.11",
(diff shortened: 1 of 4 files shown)


jacekkolasa:
yes, there was an issue that after installation the notification was shown with "Connected to the Cloud!" text. Also, @cakrit was right that we shouldn't risk unnecessary bringing notifications into non-aclk release (my bad btw.), so i'll just reenable those notifications on `86-nav-bars-release` branch.

PR #59, opened 2020-04-03
pre-release-fixes-4

jacekkolasa:
fixes bugs:
- Sign in button sometimes disappears after signing out.
- scrollbar in space-panel (sticky bottom part)

### DIFF
--- a/src/components/space-panel/space-panel.tsx
+++ b/src/components/space-panel/space-panel.tsx
@@ -80,18 +80,20 @@ export const SpacePanel = ({
       {isSignedIn
         ? <SpacePanelIframe cloudBaseURL={cloudBaseURL} streamedHostsData={streamedHostsData} />
         : (
-          <S.ScrollContainer>
-            {hasStreamedHosts && (
-              <S.PanelSection leading>
-                <ReplicatedNodes chartsMetadata={chartsMetadata} />
-              </S.PanelSection>
-            )}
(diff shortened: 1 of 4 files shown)


src/components/space-panel/space-panel.tsx
```diff
@@ -80,18 +80,20 @@ export const SpacePanel = ({
...
-            )}
-            {(machinesArray.length > 0) && (
-              <S.PanelSection leading={!hasStreamedHosts}>
-                <VisitedNodes machinesArray={machinesArray} />
-              </S.PanelSection>
-            )}
+          <>
+            <S.ScrollContainer>
+              {hasStreamedHosts && (
+                <S.PanelSection leading>
```
builat, line 86: Maybe we could create component with a prop `isVisible` to use react optimizations?

jacekkolasa: That sounds interesting, can you briefly say what it is, or paste a link? You mean to render the component always, and if it's not needed, then it should return `null`, and it is then faster?

If yes, it's interesting idea and there would be places where it could come handy (ie. on chart level for example), but here we only render it once so it doesn't make that much difference

builat: > ou mean to render the component always, and if it's not needed, then it should return null, and it is then faster?


@jacekkolasa Yes, thats was the idea

PR #63, opened 2020-04-06
force time window - prevent negative padding

jacekkolasa:
don't do unnecessary corrections when requested window is only a little different from available one

### DIFF
--- a/src/domains/chart/components/lib-charts/sparkline-chart.tsx
+++ b/src/domains/chart/components/lib-charts/sparkline-chart.tsx
@@ -30,7 +30,14 @@ const getForceTimeWindowCorrection = (
 
   const currentDuration = before - after
   const requestedDuration = requestedBefore - requestedAfter
-  const widthRatio = (currentDuration / requestedDuration)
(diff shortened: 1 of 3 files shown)


src/domains/chart/components/lib-charts/sparkline-chart.tsx
```diff
@@ -30,7 +30,14 @@ const getForceTimeWindowCorrection = (
 
   const currentDuration = before - after
   const requestedDuration = requestedBefore - requestedAfter
-  const widthRatio = (currentDuration / requestedDuration)
+  // don't do overrides when current (available) duration is bigger or only slightly lower
+  // than requested duration
+  const DURATION_CHANGE_TOLERANCE = 1.03
+  if (currentDuration > requestedDuration / DURATION_CHANGE_TOLERANCE) {
```
builat, line 36: and requestDuration never ever could be zero?

jacekkolasa: no, never, it could mean setting `after=0` and `before=0`

PR #65, opened 2020-04-09
netdata-1-21-fix-extended

jacekkolasa:
This is an fix-extension. I'm putting it in separate PR, because it's a bigger change and until we decide to fix the stable release, it's good to have both versions available.

This will go as v0.4.14 to `netdata`

Change the way we freeze elements. Clone it (so it's decoupled from React), but don't store the clone in state (to prevent updates), instead keep in ref for instant access. Update the clone's visibility based on Intersection Observer, changing the visibility immediately when it's in the scope. But note - the chart is starting to render after some async timeout (shouldHideDebounced property)


cosmix:
@jacekkolasa be advised that the code found herein was referenced in https://github.com/netdata/netdata/pull/8664 which was merged into `netdata/netdata` and released as part of the 1.21.1 hotfix release.

PR #67, opened 2020-04-13
Expose version control component

VLegakis:
The dashboard uses already a Version Control component making all the necessary calls and handling which I expose with this PR since we can easily reuse on the cloud for the required UI 
https://www.figma.com/file/OyGGjJd0Dm7GnscRvSP3Fk/%E2%9C%85S3.3---Node-View?node-id=0%3A1
https://github.com/netdata/product/issues/693
Overall they will look exactly the same as in the agent Nav bars. So better avoid any duplicate code and handling by exposing this component and reusing it in the cloud.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "0.4.14",
+  "version": "0.4.15",
   "main": "./lib/src/index-npm.js",
   "homepage": ".",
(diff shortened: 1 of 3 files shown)


jacekkolasa:
@VLegakis Yes, i think we can definitely start using this branch as a source for `@netdata/dashboard` npm package

PR #72, opened 2020-04-16
optional chaining added to unsafe call.

builat:
- relevant issue: https://github.com/netdata/product/issues/705

### DIFF
--- a/src/domains/chart/components/chart-with-loader.tsx
+++ b/src/domains/chart/components/chart-with-loader.tsx
@@ -93,7 +93,9 @@ export const ChartWithLoader = ({
 
   // (isRemotelyControlled === false) only during globalPanAndZoom, when chart is panAndZoomMaster
   // and when no toolbox is used at that time
-  const isRemotelyControlled = !panAndZoom || !isPanAndZoomMaster || shouldForceTimeRange // used when zooming/shifting in toolbox
+
+  // used when zooming/shifting in toolbox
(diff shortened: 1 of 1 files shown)


jacekkolasa:
That change looks good to me! But can we turn off prettier for Dashboard? At least for now... It is incompatible with some airbnb rules we agreed on, so i think we would have first to fix that... Applying prettier now will lead to tons of changes across the codebase and will make diffing between `master` and `nav-bars-release` branches much harder (conflicts, etc.)

src/domains/chart/components/chart-with-loader.tsx
```diff
@@ -207,47 +189,66 @@ export const ChartWithLoader = ({
...
+    shouldUsePanAndZoomPadding,
+    shouldFetch,
+    cancelTokenSource,
+  ])
 
   const [selectedDimensions, setSelectedDimensions] = useState<string[]>([])
 
-  const hasEmptyData = (chartData as DygraphData | D3pieChartData | null)?.result.data?.length === 0
+  const hasEmptyData = (chartData as DygraphData | D3pieChartData | null)?.result?.data?.length === 0
     || (chartData as EasyPieChartData | null)?.result.length === 0
```
jacekkolasa, line 252:
```suggestion
    || (chartData as EasyPieChartData | null)?.result?.length === 0
```
if `chartData` is empty object, then this second branch will trigger an error, we need to have similar handling in the second case

jacekkolasa:
@builat with that proposed change i think it will be good, we just need to fix those eslint errors

PR #84, opened 2020-05-27
misc-fixes

jacekkolasa:
various fixes for Agent Dashboard:
- fix registry when Do Not Track is enabled
- support whitespaces in replicated-nodes urls
- add trailing slash in replicated-nodes links
- disable loop to check for ACLK status (check just once, as in design)
- remove 1 level of nesting in Dashboard bundle (instead of lazy-loading, make alias in webpack based on env variable)
- add separate task to work with node-view
- describe available tasks, clean CRA readme

### DIFF
--- a/src/node-view-app.tsx
+++ b/src/node-view-app.tsx
@@ -54,7 +54,7 @@ const NodeViewApp = () => { // eslint-disable-line arrow-body-style
       })
   }, [])
 
-  const [currentRoute, setCurrentRoute] = useState(() => getCurrentRouteFromHash())
+  const [currentRoute, setCurrentRoute] = useState(getCurrentRouteFromHash)
 
   return (
     <div className="App" style={{ height: "100vh" }}>
--- a/src/node-view-app.tsx
+++ b/src/node-view-app.tsx
@@ -1,0 +1,78 @@
+import React, { useEffect, useState } from "react"
+import Ps from "perfect-scrollbar"
+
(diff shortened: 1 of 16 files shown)


src/node-view-app.tsx
```diff
@@ -0,0 +1,78 @@
...
+  useEffect(() => {
+    axiosInstance.get("http://localhost:19999/api/v1/charts")
+      .then((response) => {
+        if (response.data) {
+          setRealMetadata(response.data)
+        }
+      })
+  }, [])
+
+  const [currentRoute, setCurrentRoute] = useState(() => getCurrentRouteFromHash())
```
builat, line 57:
```suggestion
  const [currentRoute, setCurrentRoute] = useState(getCurrentRouteFromHash)
```

src/node-view-app.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <div className="App" style={{ height: "100vh" }}>
```
builat, line 60: let's declare style object out of `return` and component, so we can avoid redundant object recreation

src/node-view-app.tsx
jacekkolasa: yep

src/node-view-app.tsx
jacekkolasa: sure, done

PR #93, opened 2020-06-24
Invalidate chart upon timeframe changes

VLegakis:


src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -831,21 +831,29 @@ export const DygraphChart = ({
...
+
       // in old dashboard, when chart needed to reset internal dateWindow state,
       // dateWindow was set to null, and new dygraph got the new dateWindow from results.
       // this caused small unsync between dateWindow of parent (master) and child charts
       // i also detected that forceDateWindow timestamps have slightly better performance (10%)
       // so if the chart needs to change local dateWindow, we'll always use timestamps instead of
       // null.
 
-      const xAxisRange = dygraphInstance.current.xAxisRange()
+      const xAxisRange = dygraphInstance.current?.xAxisRange()
```
jacekkolasa, line 848: why is that question mark here? It wasn't needed before

src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -831,21 +831,29 @@ export const DygraphChart = ({
       // todo support state.tmp.dygraph_force_zoom
       const forceDateWindow = [viewAfter, viewBefore]
 
+      const dygraphOptionInEffect = dygraphInstance.current.getOption("dateWindow")
```
jacekkolasa, line 834: does `getOption("dateWindow")` give different results than `xAxisRange()` below? It's not clear why different method is used

src/domains/chart/components/lib-charts/dygraph-chart.tsx
VLegakis: this was introduced by the linter somehow. Fixed

src/domains/chart/components/lib-charts/dygraph-chart.tsx
VLegakis: Didn't realise this was the same value. Fixed

VLegakis:
> hasChangedDuration

@jacekkolasa I have change the naming. I'm not sure I understand adding this to the `hasScrolledToTheFutureDuringPlayMode` condition though. Would you like to discuss it in a short call?

VLegakis:
@jacekkolasa 
There are 2 things I worry about and wanted to think-discuss:
1) I have chosen an arbitrary threshold of 5000 ms above which we regard the duration (timeframe) has changed and request a reset. Not sure this is correct.
2) The duration (timeframe) also changes in zooming scenarios. However in these case a reset doesn't cause any issues as far as I've seen since zooming causes a reset either way.

jacekkolasa:
@VLegakis

1. 5000 is not precise, i agree, it could be 1000 too, but in this case we are just narrowing down the conditions when we detect panning to the future, so there's no harm here, i think.
2. If you zoom in/out, then our start/end times change in different directions, so it doesn't pass the rare `hasScrolledToTheFutureDuringPlayMode` criteria. 

Re: your previous question, I just wanted to have still valid value for `hasScrolledToTheFutureDuringPlayMode`. If duration has changed, then `hasScrolledToTheFutureDuringPlayMode` needs to be false, because user didn't scrolled to the future during play mode :) Does it make sense?

PR #102, opened 2020-07-21
Time selection

VLegakis:
This provides a solution for sharing with cloud app the timeframe selection.
Overall there are a lot of solutions:
1) Hash params can be manipulated by the component (this is the solution provided with this PR!)
2) Cloud-app could be listening to the setGlobalChartUnderlayAction of the Dashboard component. This proved a bit more difficult than I thought. We should think about this a bit more though because we'll run into problems requiring that sooner or later!
3) Pass in a callback to the NodeView which will be called back by the Chart. This is straightforward but requires a lot of code changes to pass this callback all across the hierarchy from the NoeView to the Charts...

Let think about these. Easiest is 1 but if we find a way to solve 2 it will help us in other issues we'll be facing soon I think

### DIFF
--- a/src/domains/dashboard/sagas.ts
+++ b/src/domains/dashboard/sagas.ts
@@ -1,3 +1,4 @@
+/* eslint-disable operator-linebreak */
 import { take, takeEvery } from "redux-saga/effects"
 import { Action } from "redux-act"
 
@@ -10,8 +11,11 @@ import {
   setGlobalPanAndZoomAction,
 } from "domains/global/actions"
 import {
-  explicitlySignInAction, showSignInModalAction, ShowSignInModalAction,
+  explicitlySignInAction,
+  showSignInModalAction,
+  ShowSignInModalAction,
 } from "domains/dashboard/actions"
+import { setHashParams, getHashParams } from "utils/hash-utils"
 
 export const LOCAL_STORAGE_NEEDS_SYNC = "LOCAL-STORAGE-NEEDS-SYNC"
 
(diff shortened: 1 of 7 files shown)


jacekkolasa:
Question is, should we manage the hash inside dashboard? On the one hand charting-library technically shouldn't manage the URL, but on the other hand we could reuse the logic behind keeping the state in url. @netdata/cloud-fe what do you think?

src/utils/hash-utils.ts
```diff
@@ -0,0 +1,52 @@
...
+  const allParamsResult = mergeDeepLeft(params, allParams)
+  // window.history.replaceState(null, "", `#${makeHashFromObject(allParams)}`)
+  window.location.hash = `${makeHashFromObject(allParamsResult)}`
+}
+
+export const getHashParam = (param: string): string => getHashParams()[param] || ""
+
+export const removeHashParams = (params: string[]) => {
+  window.location.hash = `${getFilteredHash(params)}`
+}
```
jacekkolasa, line 52: nice work with those utility functions!

Azein: @VLegakis if those are the same we use in cloud, maybe we can package them.

src/domains/dashboard/sagas.ts
```diff
@@ -32,16 +36,26 @@ function resetGlobalPanAndZoomSaga() {
...
-    const { after, before } = payload
     // additional check to prevent loop, after setting initial state from url
     if (window.urlOptions.after !== after || window.urlOptions.before !== before) {
       window.urlOptions.netdataHighlightCallback(true, after, before)
     }
+  } else {
+    // TODO: Consider a setting to control wether the component sets these hash params
+    const hashParams = getHashParams()
+    const highlightAfter = new Date(Math.round(after)).toJSON()
+    const highlightBefore = new Date(Math.round(before)).toJSON()
```
jacekkolasa, line 49: nice idea, this format will be much friendlier than timestamps. Perhaps we should also remove microseconds from the hash string?

src/utils/hash-utils.ts
VLegakis: I introduced many methods in the utils but one thing to consider is formatting.
Dashboard (on the agent) follows some patterns while we follow some others. I think it would be wise to follow the agent's hash formatting patterns. This may need some changes on our side e.g for login screens etc...
Would be good to share the same patterns I think.
About packaging, package them where?

VLegakis:
> Question is, should we manage the hash inside dashboard? On the one hand charting-library technically shouldn't manage the URL, but on the other hand we could reuse the logic behind keeping the state in url (share that logic with dashboard app). @netdata/cloud-fe what do you think?

I think it's easy and handy for the component to manipulate the hash params. Its not destructive and the utilities I provided just add/remove their params retaining any other params (so it does not interfere with the apps possible params)

As noted also we could make this behaviour optional with a setting or a param passed in (by default the component to manipulate the params)

PR #103, opened 2020-07-28
Adds community link to updates dialog

zack-shoylev:


zack-shoylev:
Also, why/how is this hidden: https://github.com/netdata/dashboard/blob/master/public/index.html#L103

jacekkolasa:
> Also, why/how is this hidden: master/public/index.html#L103

this is actually an old part of the code and we just need to clean it/remove it, along with about 10% of the index.html/main.js files.

PR #105, opened 2020-07-30
Selected dimensions

VLegakis:


src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
```diff
@@ -281,7 +281,9 @@ export const ChartWithLoader = ({
 
   useSelector(selectSpacePanelTransitionEndIsActive)
 
-  const [selectedDimensions, setSelectedDimensions] = useState<string[]>([])
+  // eslint-disable-next-line max-len
+  const selecteddimensionsMapped = attributes?.selectedDimensions?.filter((dim) => dim.correlationScore !== undefined).map((dim) => dim.name) || []
+  const [selectedDimensions, setSelectedDimensions] = useState<string[]>(selecteddimensionsMapped)
```
jacekkolasa, line 286: does low level `chart-with-loader` needs to know about `correlationScore`? Can we have those conditionals and here use just lower level `selectedDimensions` as strings or array of strings?

jacekkolasa:
Also please be aware that if you turn off some dimensions, they will still be fetched by Dashboard

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
VLegakis: What I'm doing here with correlationScores is play safe by just returning an empty array (same as it was before) in case there is no correlation score.
another issue to have in mind here is that in a future iteration I will need to display these scores on the charts. So I'd prefer using the whole dimension instead of just an array of strings. Not 100% sure about these though...

src/domains/dashboard/components/node-view/render-submenu-name.tsx
```diff
@@ -124,6 +124,9 @@ export const renderSubmenuName = ({
                 decimalDigits: netdataDashboard.contextDecimalDigits(
                   chart.context, -1,
                 ),
+                selectedDimensions: Object.values(
+                  chartsMetadata.charts[chart.id]?.dimensions || {},
```
jacekkolasa, line 128: @VLegakis i understand that `chartMetadata` is modified in the cloud and sent to dashboard. I see few problems with that. First is that Dashboard expects that metadata to be the same as the one we fetch from BE. It looks like chartsMetadata were changed only to omit props-drilling, which IMO shouldn't happen. If we have problem with props drilling, we should perhaps go around it with putting something to redux state, or create some other configuration-like object, which is later mapped to attributes in Dashboard?

Also, the `dimensions` in chartsMetadata will not be used in the future, and it's even possible that we'll remove that support from the Dashboard. That's because the dimensions can change during chart updates (can be added dynamically by result of data calls), and we fetch metadata only once, on init time. We use `dimensions` from metadata currently only for colors, but it's actually wrong and there's a bug related to this.

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
jacekkolasa: I am trying to have dashboard working in more lower-level, so it doesn't grow that much, it's already quite complicated. For example we could have the "selectedDimensions" which can be used in many different ways (ie. for other features than correlationScores as well), and if you want to display some new elements in charts (in the future) - we can have just one prop like "customElement" (we already have `customElementForDygraph` for example)

src/domains/dashboard/components/node-view/render-submenu-name.tsx
VLegakis: Ok understood. Maybe got the `attributesOveride` way then. Lets discuss this after the demo

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
VLegakis: I removed the filtering from the array since now its part of the attributes overrides.
However I retains the Dimension[] struct (instead of array of strings) since there will be some addition info required to be shown there soon. 
More importantly it doesn't make much difference if the mapping happens on the parent component or here I think (let me know if this is the case and I'll make the mapping above it)
Please note if there are no selectedDimension it directly defaults to empty array (no mapping is done...)

src/domains/dashboard/components/node-view/render-submenu-name.tsx
VLegakis: Done

src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -69,7 +71,13 @@ const SubSection = memo(({
             .map(parseChartString)
             .map((attributes: Attributes | null) => attributes && (
               <ChartWrapper
-                attributes={{ ...attributes, host }}
+                attributes={
+                  {
+                    ...attributes,
+                    host,
+                    ...(attributesOverrides ? attributesOverrides[attributes.id] : {}),
```
builat, line 78: could it be placed in function?

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
jacekkolasa: @VLegakis the `attributes` object structure is something like our public, low level Dashboard's API. For example, people can put those attributes as `data-something=...` DOM attributes, in static custom dashboards (html files). So i think this `attributes` object should stay as much abstract as possible. We should also document in the future those settings in `https://learn.netdata.cloud/docs/agent/web/gui/custom#selecting-dimensions` page. If we'll have the new structure, we will need to write in docs:
> If you need to have some dimensions pre-selected, add selected dimensions in following format: `data-selected-dimensions="[{name: dimension1, correlationScore: true},...]"`.

Even is someone will not use `correlationScore` at all. That can go out of hand. I understand that we'll need some additional features for `correlationScore`, like displaying some additional info, elements, etc. but for that we could have some additional api (additional properties in attributes), even as reactNode type.

But I also don't want to block you, and I understand that going with the current implementation could be quicker, so if you want to merge it and move forward, please go ahead, we can eventually change it in the future.

VLegakis: Got it! I keep on forgetting this aspect of dashboard component. I'll provide a new PR with simple array of strings. We'll see what we'll do once scores are also required there for rendering

VLegakis: Done. So, now `Attributes` are like:
selectedDimensions: ["dim1", "dim2", ...] etc
and also 
data-selected-dimensions: ["dim1", "dim2", ...] etc

src/domains/dashboard/components/node-view/node-view.tsx
VLegakis: Could be, I just followed @jacekkolasa 's approach and decided to stay consistent:
```
 // add commonMin/commonMax attributes only if they are set
                ...(commonMin ? { commonMin } : {}),
                ...(commonMax ? { commonMax } : {}),
                ...(attributesOverrides ? attributesOverrides[chart.id] : {}),
```

src/domains/chart/chart-types.ts
```diff
@@ -42,8 +42,9 @@ export interface EasyPieChartData extends ChartDataBase {
 
 export type ChartData = DygraphData | EasyPieChartData | D3pieChartData
 
-interface Dimension {
+export interface Dimension {
   name: string
+  correlationScore?: Number
```
VLegakis, line 47: I'm leaving this to be optionally used in charts metadata. (Attribute use just a string[])

VLegakis:
Known issue to handle:
Chart handles selectedDimensions either externally or manually. Meaning if you pass-in selectedDimensions attributes from parent component, a user cannot manually select-deselect dimensions. 
We can improve this by supporting both "modes" in a future iteration. This requires a bit more thought

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
jacekkolasa: Thanks @VLegakis!

PR #115, opened 2020-08-25
y-axis and scaling for constant values fix

jacekkolasa:
change dygraph version to the same which was used in old dashboard. That version had a fix from @ktsaou for constant other than 0. In those cases no y-legend was present and autoscaling was broken.

Instead of uploading full js file to the repo, i linked already existing file in /public, which normally is used for hosting old dashboard in the Agent. I checked that `npm run build` works fine and fixes similar problem when added to cloud-fe.

related issue: https://github.com/netdata/netdata/issues/9659

old issue explaining the problem: https://github.com/netdata/netdata/pull/3407

### DIFF
--- a/src/domains/chart/components/lib-charts/dygraph-chart.tsx
+++ b/src/domains/chart/components/lib-charts/dygraph-chart.tsx
@@ -3,7 +3,9 @@ import React, {
   useLayoutEffect, useRef, useCallback,
 } from "react"
 import { useUpdateEffect, useUnmount, useMount } from "react-use"
-import Dygraph from "dygraphs"
+// this version is needed because it contains a fix for handling constant value in the chart
+// ie. https://github.com/danvk/dygraphs/pull/909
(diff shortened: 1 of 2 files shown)


VLegakis (approved):
Ok we can revert to the old version but shouldn't we fix this in the latest dygraph version too?
Let me know if I can help and/or do it together @jacekkolasa

VLegakis:
Of course we can proceed with this until we permanently solve it in the latest dygraph version

jacekkolasa:
@VLegakis i don't think it's possible to push for new dygraph version now (i mean this repo https://github.com/danvk/dygraphs ). I think what @ktsaou did is he locally added a change from [this pr](https://github.com/danvk/dygraphs/pull/909), build a minified dygraph version and he added it to old dashboard. So actually the npm version of dygraph was older (didn't had Costa's fix)

PR #116, opened 2020-08-26
Allow manually selected dimensions override externally selected dimensions

VLegakis:
This allow manually selected dimensions override Externally Selected Dimension


novykh (approved):

jacekkolasa (approved):
lgtm! the audits most likely come from upgraded react-script

PR #120, opened 2020-08-28
Menu counters for metrics correlations

VLegakis:
PR's result with menu counters:
Screenshot 2020-08-27 at 9 14 47 PM

### DIFF
--- a/src/domains/dashboard/components/node-view/node-view.tsx
+++ b/src/domains/dashboard/components/node-view/node-view.tsx
@@ -110,6 +110,7 @@ interface Props {
   scrollableContainerRef: React.RefObject<HTMLDivElement>
   timeWindow?: number
   attributes?: ChartsAttributes
+  metricsCorrelationMetadata?: ChartsMetadata
 }
 export const NodeView = ({
   chartsMetadata,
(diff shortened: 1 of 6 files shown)


src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -152,7 +154,9 @@ export const NodeView = ({
     ((((width * pcentWidth) / 100) * chartsMetadata.update_every) / 3) / 60,
   ) * 60
 
-  const menus = useMemo(() => renderChartsAndMenu(chartsMetadata), [chartsMetadata])
+  const menus = useMemo(() => renderChartsAndMenu(metricsCorrelationMetadata
+    || chartsMetadata, metricsCorrelationMetadata
+    && chartsMetadata), [metricsCorrelationMetadata, chartsMetadata])
```
jacekkolasa, line 159: do you think that destructuring named arguments (ie. `renderChartsAndMenu({ chartsMetadata, metricsCorrelationMetadata })` would be more readable?

jacekkolasa (commented):
good stuff, i understand it's work in progress, right? (draft pr).

src/domains/dashboard/components/node-view/node-view.tsx
VLegakis: Cleaned up a bit this part. Please retake a look

jacekkolasa (approved):
lgtm! IMO it would be good someday to split ChartsMetadata from Metrics Correlation state (which are not metadata, ie. don't come from /charts endpoint). But right now it works, and from what i see it will be easy to clean it in the future.

PR #121, opened 2020-09-02
fill fake nulls when requesting more than available history

jacekkolasa:
When requesting for bigger time interval than available history in the agent, we get only the available range. Dashboard was first designed to not allow zooming-out too much, so when we were requesting too big range, we were getting in response the same nr of points requested, but in smaller time frame (for available history). We want to force the requested time frame, so that bigger frequency is a waste, and instead we'll now request reduced nr of points, and after getting response, we'll fill empty data, so that charts have requested time-frame on x-axes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "1.4.3",
+  "version": "1.5.0",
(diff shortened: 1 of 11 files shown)


VLegakis (approved):
Many thanks for this! Great job :)

PR #122, opened 2020-09-03
Used css vars for easier manipulation from external renderer.

novykh:
Using css variables on the :root selector, makes it easy to change/override those variables from the cloud app and provide colors based on the cloud theme.


builat (approved):

jacekkolasa (approved):

jacekkolasa (commented):
@novykh please update also package.json version, to track releases properly

PR #123, opened 2020-09-09
New themes

novykh:
Added style vars for default theme css files.

- Will be overridden from DashboardThemeProvider when used, otherwise fallback to default values.

### DIFF
--- a/src/components/app-header/components/alarms-control/alarms-control.tsx
+++ b/src/components/app-header/components/alarms-control/alarms-control.tsx
@@ -1,8 +1,7 @@
 import React from "react"
+import { Button } from "@netdata/netdata-ui"
 import { IconContainer } from "../../styled"
-import {
-  StyledButton, IndicatorsContainer, AlarmIndicator, StyledButtonContainer,
-} from "./styled"
+import { IndicatorsContainer, AlarmIndicator } from "./styled"
 
 interface Props {
   criticalAlarmsCount: number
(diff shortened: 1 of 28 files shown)


src/components/app-header/components/alarms-control/alarms-control.tsx
```diff
@@ -17,13 +16,16 @@ export const AlarmsControl = ({
       {criticalAlarmsCount > 0 && <AlarmIndicator alarmType="critical" />}
       {warningAlarmsCount > 0 && <AlarmIndicator alarmType="warning" />}
     </IndicatorsContainer>
-    <StyledButtonContainer
-      href="#"
+
+    <Button
+      flavour="borderless"
+      neutral
+      themeType="dark"
```
Azein, line 23: What is themeType?

VLegakis (commented):
Just wanted to understand what does this PR affect?
Does it affect only the NodeView or it will affect the open source Agent Dashboard too?
I've also seen we are using Styled components. Which components is this targeting?

src/components/app-header/components/alarms-control/alarms-control.tsx
novykh: Please refer to the [discussion](https://github.com/netdata/netdata-ui/pull/110#discussion_r479283022) having about this on the ui kit. If it's not clear we can discuss more.

novykh:
> Just wanted to understand what does this PR affect?
> Does it affect only the NodeView or it will affect the open source Agent Dashboard too?
> I've also seen we are using Styled components. Which components is this targeting?

Well it's about the node view only. Meaning only for the cloud (If it's used anywhere else, please let me know).
It's also getting rendered on the `netdata/dashboard/src/node-view-app.tsx` which is just for local testing I guess (again if it's used for other cases on production, please let me know)

The main change are the css vars, that when provided will update the default (white) theme of the dashboard. But, dashboard itself doesn't use the css vars, thus will remain untouched by this PR. Whereas on the cloud, where we do not have any way to change the theme dynamically, we are passing those css vars, through the theme Prop on the `DashboardThemeProvider`.

VLegakis (approved):
Sound good Johnny. If the above are correct (about affecting only the Node View) we are good to go with this one. Just wanted us to be extra careful with the Dashboard since it affects the OS agent too some times in unpredictable ways. Jacek is missing also to double check. 
Before Merging these to master maybe we should be testing the agent dashboard too making sure nothing breaks there

src/components/app-header/components/alarms-control/alarms-control.tsx
Azein: So, now the Button component should be aware of what themes exactly do exist, to correctly present itself?

novykh: Nope, that's just a way to provide custom styles (in that case styles from a different theme) without the need to write the styles yourself.

PR #126, opened 2020-09-15
Updated charts and menu to use chart name instead of type attr.

novykh:


builat (approved):

src/domains/dashboard/utils/render-charts-and-menu.ts
```diff
@@ -9,11 +9,13 @@ import { netdataDashboard, options } from "./netdata-dashboard"
 // to reflect our menu system and content
 // TODO: this is a shame - we should fix charts naming (issue #807)
 // ^^ (original comment from gsmox, still valid!)
-function enrichChartData(chart: ChartMetadata) {
-  const parts = chart.type.split("_")
+function enrichChartData(chartName: string, chart: ChartMetadata) {
```
VLegakis, line 12: I'm not sure about that. Shouldn't it be safer to have a separate function doing this for the Room Overview and leave this here for the NodeView and Agent Dashboard?

VLegakis (changes requested):
I think it'd be safer leaving the existing charts enrichment functionality independent.
Except if we can guarantee the same function can be used for all cases. But this can prove difficult to maintain. Lets discuss about this

src/domains/dashboard/utils/render-charts-and-menu.ts
novykh: There is no reason to duplicate code in this case. The type is safely extracted from the name on both node and context charts.

VLegakis: Ok lets double check it can be reused for all cases (Overview, NodeView, AgentDashboard)
Something else: why change the function signature to include chartName while chart name is included in ChartMetadata?

VLegakis (changes requested):
Ok lets double check it can be reused for all cases (Overview, NodeView, AgentDashboard)
Something else: why change the function signature to include chartName while chart name is included in ChartMetadata?

VLegakis (approved):

VLegakis:
@novykh what else may be missing is the NodeView params.
1) I'm not sure if it makes sense to create a new RoomOverview component in addition to NodeView. We'd better reuse-enhance Nodeview but if no other option...
2) In case we reuse NodeView note sure the chartMetadata params of the component passed in will need changes...

src/domains/dashboard/utils/render-charts-and-menu.ts
VLegakis: Also, shouldn't it be something other than ChartMetadata for the RoomOverview case? (If this is true it brings back the question shouldn't it be better a new separate function?)

VLegakis (changes requested):
Lets not merge this until we finalise a bit better the data types please (Chart Metadata may be deprecated for the new endpoint)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "1.5.4",
+  "version": "2.0.0-alpha.7",
```
jacekkolasa, line 3: if this is `alpha`, perhaps we should have separate branch for this? `master` is used to release new versions for the Agent

src/dashboard.js
```diff
@@ -7488,6 +7488,10 @@ let chartState = function (element, chartIndex) {
         if (this.dimensions) {
             this.data_url += "&dimensions=" + this.dimensions;
         }
+                
+        if (this.aggr_method) {
+            this.data_url += "&aggr_method=" + this.aggr_method;
+        }
```
jacekkolasa, line 7494: It doesn't look like it's intentional, is it a mistake? There is no `aggr_method` inside dashboard.js. To be clear, this file is used only for static custom dashboards in the Agent (like tv.html, etc.)

(same goes for `public/dashboard.js` file)

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -553,3 +558,62 @@ export const getAttributes = (node: Element): Attributes => {
...
+    watt: true,
+    temperature: true,
+    "random number": true,
+    rpm: true,
+    quadro: true,
+    "adv/item": true,
+    multiplier: true,
+    geforce: true,
+  }
+  if (avgUnits[unit.toLowerCase()]) {
```
jacekkolasa, line 602: `avgUnits` could be also an array and then `avgUnits.includes(unit.toLowerCase())`, but that's very minor :)

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -553,3 +558,62 @@ export const getAttributes = (node: Element): Attributes => {
   const showValueOf = getAttributesDynamic(node)
   return { ...attributesStatic, showValueOf }
 }
+
+export const mapDefaultAggrMethod = (unit: string): string => {
```
jacekkolasa, line 562: `transformDataAttributes` file is used for functions parsing DOM attributes for charts created with `data-netdata=".."` DOM attributes, perhaps it could be put just in some utils file?

src/utils/fill-missing-data.ts
```diff
@@ -60,3 +60,22 @@ export const fillMissingData = (data: ChartData, nrOfPointsToFill: number) => {
...
 }
+
+const emptyArray: number[] = []
+export const transformResults = (data: ChartData, format: string) => {
+  if (format === "array" && data.format === "json") {
+    if (Array.isArray(data.result)) return data
+
+    return {
+      ...data,
+      result: (data.result.data as number[][]).reduce((acc: number[], pointData: number[]) => {
```
jacekkolasa, line 71:
```suggestion
      result: ((data as DygraphData).result.data).reduce((acc: number[], pointData: number[]) => {
```
that could be cleaner typing, or at least similar to current pattern (ChartData is an union of different *Data types), this way we know what type of data is coming from BE

src/domains/chart/utils/transformDataAttributes.ts
novykh: There is a small perf boost, since it avoids looping through the array - I suggest to leave it as is.

package.json
novykh: Yep, won't be merged until it reaches a stable version.

src/domains/chart/utils/transformDataAttributes.ts
jacekkolasa: I'm ok with this, I wasn't aware it's a tweak for performance. But IMO (in the future) it would be good if we don't do micro-optimisations like this often - it makes code less readable, gives low value, and if we have some performance problem, I suggest we fix them on a higher level (for example with useMemo). That's just my personal preference.

Btw. optimisation would be better if `avgUnits` were created outside the function :)

novykh: https://github.com/netdata/dashboard/pull/127#discussion_r489446927

src/domains/chart/utils/transformDataAttributes.ts
VLegakis: I've put it in `utils/fill-missing-data`

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
```diff
@@ -251,6 +252,8 @@ export const ChartWithLoader = ({
           after: after || null,
           before: before || null,
           dimensions: attributes.dimensions,
+          aggrMethod: attributes.aggrMethod,
```
jacekkolasa, line 255: for choosing the aggregation method, there is an old `group` param. Do you know why `aggrMethod` is used instead of `group`?

https://learn.netdata.cloud/docs/agent/web/gui/custom#data-grouping-method

src/dashboard.js
VLegakis: I wanted to discuss about this. Removed and we can discuss it later on

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -85,6 +85,7 @@ export interface StaticAttributes {
   units?: string
   unitsCommon?: string
   unitsDesired?: string
+  aggrMethod?: string
```
jacekkolasa, line 88: can we use existing `method` attribute?

https://learn.netdata.cloud/docs/agent/web/gui/custom#data-grouping-method

src/dashboard.js
VLegakis: @jacekkolasa I wanted to ask also whether it makes sense for the aggregation methods to be set in the relevant `snapshot-loaders`

src/domains/dashboard/utils/render-charts-and-menu.ts
jacekkolasa: Thanks @VLegakis for getting attention to this. After testing, this change looks ok for me. I was a little unsure, but i've checked on our demo servers, that given chart keys, `key.split(".")[0]` always equals `chart.type`. So we can use that now. It would be good though to double-check in Agent team.

jacekkolasa: > Something else: why change the function signature to include chartName while chart name is included in ChartMetadata?

From my understanding this is a change that allows us to work also when we don't have `chart.type` property, the only place we can get it is by extracting it from `charts[key]`

src/dashboard.js
jacekkolasa: Right now we don't need to have it there, but they can stay, so we don't miss it when we'll introduce aggr methods in Agent Dashboard

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
VLegakis: This is important. So what is exactly `group` used for? 
Perhaps we need to contact @netdata/cloud-be and use that existing one

jacekkolasa: from what i understand `group` is used for the same thing, different aggregation methods :) Users even not can change that param in their custom `dashboard_info.js`

VLegakis: From the above documentation is seems that the grouping method  (`group` param) is used when the Netdata server is requested to return fewer points for a time-frame. 
aggrMethod is for aggregation across room multi nodes...

jacekkolasa: Sorry, i should have wrote `method` param. It's used for the same thing. `group` is the attribute name.

src/domains/chart/utils/transformDataAttributes.ts
VLegakis: From the above documentation is seems that the grouping method (group param) is used when the Netdata server is requested to return fewer points for a time-frame.
aggrMethod is for aggregation across room multi nodes...

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
jacekkolasa: It is used for the very same thing, perhaps we should have the same naming on BE side (ie. use `method` URL param)

jacekkolasa: We talked on slack and it turns out I misunderstood, `aggrMethod` is for choosing grouping method across nodes, whereas `method` is for grouping across time, in single node. Naming is a little confusing, but perhaps comments will be enough...

src/domains/chart/utils/transformDataAttributes.ts
jacekkolasa: @VLegakis please disregard, re: our discussion about aggregation across nodes

src/domains/chart/components/snapshot-loader.tsx
```diff
@@ -50,7 +50,7 @@ const SnapshotLoader = ({
       id: chartUuid,
     }))
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []) // todo fetch based on state
+  }, [attributes]) // todo fetch based on state
```
jacekkolasa, line 53: what is the reason behind this? I intentionally set it to run only once, is it not working properly?

VLegakis: Removed! I had it there based on the rest :
https://github.com/netdata/dashboard/pull/126#discussion_r492636656

jacekkolasa:
I'm ok with merging this, but we'll need at some point have better typings for Chart Metadata, because right now we have 3 different structures - for node-view, room-overview (some props like first_entry missing), and table-view (more props missing)

jacekkolasa (approved):
can we merge this pr and set non-alpha version?

jacekkolasa (approved):
lgtm

VLegakis (approved):

PR #127, opened 2020-09-16
Default aggr method

VLegakis:


novykh (approved):
Let's merge it.

src/domains/chart/sagas.ts
```diff
@@ -134,7 +134,7 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
     before,
     dimensions,
     context: chart,
-    aggr_method: "avg",
+    aggr_method,
```
novykh, line 137: We should keep `aggrMethod` attribute as camelCased on every place inside the code, and just pass it here as `underscored`
```
aggr_method: aggrMethod,
```

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -318,6 +320,7 @@ const getAttributesMap = (): AttributesMap => ({
   units: { key: "units" },
   unitsCommon: { key: "common-units" },
   unitsDesired: { key: "desired-units" },
+  aggr_method: { key: "aggr-method", defaultValue: "avg" },
```
novykh, line 323: You should use the `mapDefaultAggrMethod` here.

src/domains/chart/utils/transformDataAttributes.ts
```diff
@@ -540,10 +543,54 @@ export const getAttributesStatic = (node: Element): Attributes => mapObjIndexed(
   getAttributesMap(),
 ) as Attributes // need to override because of broken Ramda typings
 
+const mapDefaultAggrMethod = (unit: string): string => {
+  const avgUnits = [
```
novykh, line 547: Let's put this outside of the function itself (as a constant) and transform it to an object:
```
{
   percentage: true,
// or
  percent: "percent",
```
That way the complexity will be O1, since this:
```
avgUnits.some((avgUnit) => unit.includes(avgUnit)
```
will be just 
```
!!avgUnits[unit]
```

novykh (changes requested):
Let's make those changes and merge it

src/domains/chart/utils/transformDataAttributes.ts
VLegakis: This is not possible to be used here

novykh: OK, great, but we need to pass a default value based on the chart unit. Where do you think we should put this kind of logic?

src/domains/chart/utils/transformDataAttributes.ts
VLegakis: Done

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
```diff
@@ -236,7 +236,7 @@ export const ChartWithLoader = ({
       }) : null
 
       const group = attributes.method || window.NETDATA.chartDefaults.method
-
+      console.log("attributes", attributes)
```
novykh, line 239: please remove.

novykh (approved):

PR #128, opened 2020-09-17
aggrMethod regexes

VLegakis:


novykh (approved):

cakrit (changes requested):
nice. the only thing I saw missing is "celsius" (lower case C).

VLegakis:
> nice. the only thing I saw missing is "celsius" (lower case C).

That reminded me the lower cases. I check always against lowercase now. Can you check if it looks ok?

cakrit (approved):

PR #129, opened 2020-09-23
time-frame bar

jacekkolasa:
- sticky time-frame bar for Agent Dashboard
- display date instead of 0:00 time on x-axis (useful when viewing with long history)


VLegakis (commented):
This looks good but I'm a bit confused about its purpose.
I see there is a time selection component introduced here?
Isn't this external (e.g on the cloud) which is just passing the attributes in the dashboard.
At least this is how we did it until now for NodeView and TableView as far as I remember...

builat (approved):

jacekkolasa:
@VLegakis The time selection component is added here only for Agent Dashboard. It will be replaced by date-range picker in the future, which @builat works on. I had to make a different way of passing the attributes (attributes.after to be exact), so that it resets the live mode after you do panning and then click on the time-frame buttons. Also, Agent Dashboard doesn't update the charts after they are created. It was easier to just have that state directly in Dashboard. I can add the same behaviour in the Cloud later, and expose that frame-selection state as react hooks. What do you think?

VLegakis (approved):
Great! Thanks for the clarifications @jacekkolasa !

PR #130, opened 2020-09-25
legend on bottom

jacekkolasa:
https://github.com/netdata/dashboard/pull/126 needs to be merged first

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.0.0-alpha.10",
+  "version": "2.0.0-beta.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.0.0-beta.0",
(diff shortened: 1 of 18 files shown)


VLegakis (commented):
That looks good. Only thing here is that its difficult to review since it includes all composite-charts branch changes. Shouldn't this PR be pointing toward this branch?

jacekkolasa:
@VLegakis i did it this way because i thought that we'll merge composite-charts branch anyway soon. I can change the target branch, but i don't understand why we need to keep composite-charts not merged

VLegakis (approved):

xPaw:
I'm not really a particular fan of having the legend on the bottom, especially on wide screens, where time is on right, and legend is on the left. When moving mouse around on the chart to see metrics at a specific time, I keep needing to glance at opposite sides of the screen.

Is there no user facing option to toggle it back to have legend on the right, instead of bottom?

jacekkolasa:
@xPaw thanks for your input. Indeed - comparing dimensions is harder in this view... We'll add a configuration option for it, hopefully we'll have time to do it before next stable release.

cakrit:
Giving an option is correct, but the issue identified here is important. We need to improve this. 
e.g.
-  Align the dimensions to the right instead of left.
- Show a legend on mouseover.

PR #131, opened 2020-09-28
fix node-view memoization when there are no nodeIDs

jacekkolasa:


novykh (approved):

VLegakis (approved):
Nice catch! We should be looking more into such issues

PR #132, opened 2020-09-29
empty node-ids fix

jacekkolasa:
https://github.com/netdata/product/issues/1441

### DIFF
--- a/src/domains/chart/sagas.ts
+++ b/src/domains/chart/sagas.ts
@@ -135,7 +135,7 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
     dimensions,
     context: chart,
     ...(aggrMethod && { aggr_method: aggrMethod }),
-    ...(nodeIDs && { node_ids: nodeIDs.join(",") }),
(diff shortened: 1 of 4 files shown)


novykh (approved):

src/domains/chart/sagas.ts
```diff
@@ -135,7 +135,7 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
     dimensions,
     context: chart,
     ...(aggrMethod && { aggr_method: aggrMethod }),
-    ...(nodeIDs && { node_ids: nodeIDs.join(",") }),
+    ...(nodeIDs && nodeIDs.length && { node_ids: nodeIDs.join(",") }),
```
VLegakis, line 138: Hmmm @jacekkolasa @Azein @novykh  
This means that if the filter returns no Node it will show all Nodes? Is this the expected behaviour?

jacekkolasa: i was worried about it, too, but @lvrach confirmed that no node-ids (`&node_ids=` at the end) will have the same result as not providing node_ids at all. Makes sense for me, but i agree we need to be careful here

PR #134, opened 2020-10-02
Show nodes in chart

novykh:


builat (approved):

src/domains/chart/components/chart-with-loader/chart-with-loader.tsx
```diff
@@ -50,9 +52,10 @@ import * as S from "./styled"
 import "./chart-with-loader.css"
 
 export type RenderCustomElementForDygraph = (selectedChartConfiguration: {
-  attributes: Attributes,
-  chartMetadata: ChartMetadata,
-  chartID: string,
+  attributes: Attributes
+  chartMetadata: ChartMetadata
+  chartID: string
+  getChartData: () => ChartData | null
```
VLegakis, line 58: @novykh  what is this getChartData required for?
Don't we have the nodeIDs in the attributes? (first argument here).
Am I missing something? Maybe it doesn't harm passing this also @jacekkolasa ?


Screenshot 2020-10-02 at 3 28 21 PM

jacekkolasa: i personally don't see any harm with providing that data to Cloud. From what i understand it's used as a ref to prevent some updates, correct? It would be good to know what is updated unnecessarily and if we have problems with performance here. Because that ref can get unsynced if not used correctly. Let's keep it in mind that we need to use it with caution, and we should be good...

novykh: Yep, it wont trigger unneeded rerenders. The reason i pass it as a closure is to avoid race conditions - showing old data on the cloud. Of course this is the first implementation and if we encounter any issues we will resolve them then. But I dont think we will

PR #142, opened 2020-10-09
Fixed default aggregation method for gauge charts.

novykh:


src/domains/dashboard/components/head-main.tsx
```diff
@@ -261,11 +262,11 @@ export const HeadMain = ({
                 chartLibrary: "easypiechart",
                 title: "IPv6 Inbound",
                 units: "kbps",
-                aggrMethod: "sum",
                 width: "11%",
                 before: 0,
                 unitsCommon: "system.ipv6.mainhead",
                 ...(attributesOverrides && attributesOverrides["system.ipv6"]),
+                aggrMethod: mapDefaultAggrMethod("%"),
```
VLegakis, line 269: I don't think the unit here are %

VLegakis (commented):
So how does this PR fix the issue? Doesn't it return the same aggMethods as they were?

Units were hardcoded here so was the aggrMethod. I only see some performance degradation with this PR by calling unnecessarily mapDefaultAggrMethod.
In addition to that and more importantly, we may want to override especially for these gauges the aggMethod (to set it as differenct than the default aggrMethod) so I think we'd better leave as it was (simple string).

We can just fix the `sum` to `avg` for the kbps gauge (which is also hard coded with this PR by passing in % instead of kbps...)

src/domains/dashboard/components/head-main.tsx
novykh: good catch, I've updated it

VLegakis: Many thanks @novykh ! 
However the problem is not fixed I think. I think we should be using hard coded strings.
With the latest commit you pushed, the aggMethod will return as "sum" whihc is the correct specified in the algorithm.
However I think we want "avg" here! 
So I propose we don't use the `mapDefaultAggrMethod` not only for performance reasons but more importantly because we want to hard code the aggrMethod too (like in that case)

novykh: it was hard codes as sum before. What is the issue?

VLegakis: I repeat it above in 3 comments: The way it is now it will also return "sum" which is wrong. There is no change from what it was in aggregation methods! (apart from unnecessarily calling `mapDefaultAggrMethod` which I propose it be removed)
If it's is not clear please give me a call.

VLegakis:
Provided fix here: https://github.com/netdata/dashboard/pull/143

VLegakis (approved):
Added some attributes overrides for `units` and `dimensions`. @jacekkolasa @novykh  let me know if you think of anything more required to be overridden there specifically for Head panel `Gauges`

src/domains/dashboard/components/head-main.tsx
```diff
@@ -65,13 +66,13 @@ export const HeadMain = ({
...
               ...commonAttributes,
               id: "system.io",
-              dimensions: "in",
               chartLibrary: "easypiechart",
               title: "Disk Read",
               width: "11%",
               before: 0,
               unitsCommon: "system.io.mainhead",
               ...(attributesOverrides && attributesOverrides["system.io"]),
+              dimensions: "in",
```
jacekkolasa, line 75: if we override the override, the former is no longer an override :) Can we maybe instead tweak the override in the cloud so it doesn't break default values?

VLegakis: I thought the same. So, essentially these are hard coded gauge charts which should not be changed by external attribute overrides. 
Meaning if we change any of `aggrMethod`, `units` or `dimensions` (please add anything else if I don't see it here) these Gauges should not be affected correct? So if we change the system.io's dimensions in the cloud the relevant chart should change with a new dimension BUT this gauge should remain constant like it was intended.
So, we could remove attributesOverriddes altogether. I just leave them here in case we want to override something else in the future (I don't maybe this doesn't have any meaning or usefulness and better remove it)

jacekkolasa: @VLegakis what about `chartsAttributes` which you added for correlation mode? Is this information also only applying to dygraph charts? If yes, then maybe we could just remove the attributesOverrides altogether from head-main, etc? Optionally, renaming that param to `dygraphAttributesOverrides`? We have `commonAttributesOverrides` now which we can use to pass params to *every* chart. What do you think?

VLegakis: Yes, I think this is the best approach. Metric Correlations is also using only the dygraphs. Removing them then

VLegakis: Please take a look @jacekkolasa and let me know

jacekkolasa (approved):
thanks @VLegakis , i think it's a clean solution. Of course we don't know what future will bring us, but at least it's simpler now :)

PR #144, opened 2020-10-12
Chart refresh upon attributes change

VLegakis:


novykh (approved):
Should this happen on aggr method change as well?

builat (approved):

jacekkolasa:
> Should this happen on aggr method change as well?

i think it should!

jacekkolasa (approved):

VLegakis:
Correct! I thought we did that already for the aggrMethod. Adding it now

PR #146, opened 2020-10-13
Update aggrMethods and units

VLegakis:
So there is an issue regarding the Guages: 
We want them not getting affected by aggregation methods and dimensions coming in from the cloud (So if someone plays around with the relevant chart's definition the top Gauges are not affected)
In this spirit I sent the last PR ignoring the attribute overrides coming from the cloud specifically for these charts as we agreed with @jacekkolasa .
BUT this means the aggrMethod param will always be added in the gauge Charts data calls. 
Meaning even for Agent Dashboard.
It doesn't cause any issue for the moment since agents just ignore this param. 
However I think we should find some better solution.
For the moment I add anywhere it was missing the aggMethods and relevant units TBD

### DIFF
--- a/src/domains/dashboard/components/head-main.tsx
+++ b/src/domains/dashboard/components/head-main.tsx
@@ -69,6 +69,7 @@ export const HeadMain = ({
               width: "11%",
               before: 0,
               unitsCommon: "system.io.mainhead",
+              aggrMethod: mapDefaultAggrMethod("kbps"),
             }}
             style={{ marginRight: 10 }}
             chartMetadata={chartsMetadata.charts["system.io"]}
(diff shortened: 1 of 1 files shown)


builat (approved):

jacekkolasa:
@VLegakis maybe in that case in the future we should have an explicit param to control it? Right now, if there is no `httpMethod === "POST"` in commonAttributes, then `aggrMethod` is ignored. I understand it's not intuitive, but does it solve your concern?

VLegakis:
Ok I see @jacekkolasa . Thanks for that. We can go with it (this explicit POST check) and think of something better in the future (like the flag you mention) So I suppose this PR should be fine for now. 
Please check also if removal of `units:` since it also has `commonUnits:` is Ok and we are good to go.

PR #148, opened 2020-10-13
composite-charts-heads-fix

jacekkolasa:
fix heads and mainheads to use default aggrMethod (separate from selected option for dygraphs in the cloud)
fix GET calls instead of composite-chart POST for heads

### DIFF
--- a/src/domains/dashboard/components/node-view/node-view.tsx
+++ b/src/domains/dashboard/components/node-view/node-view.tsx
@@ -3,6 +3,7 @@ import React, {
 } from "react"
 import { useScroll } from "react-use"
 import { name2id } from "utils/name-2-id"
+import { mapDefaultAggrMethod } from "utils/fill-missing-data"
 import { ChartsMetadata } from "domains/global/types"
 import { Attributes, ChartsAttributes } from "domains/chart/utils/transformDataAttributes"
 import { DropdownMenu } from "domains/chart/components/chart-dropdown"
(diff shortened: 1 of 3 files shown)


src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -81,10 +82,10 @@ const SubSection = memo(({
                   {
                     ...attributes,
                     forceTimeWindow: true, // respect timeWindow
+                    aggrMethod: mapDefaultAggrMethod(chartsMetadata.charts[attributes.id].units),
```
VLegakis, line 85: Nice! But how would we be able to pass in attributes from the cloud now?
That may be because I haven't followed the commonAttributes development. 
So what is their difference with the attributesOverrides?

jacekkolasa: Thanks for asking! It's related to what we discussed [here](https://github.com/netdata/dashboard/pull/142#discussion_r503310322)? That is - `commonAttributesOverrides` , with attributes for specific chart ids should now be pointed only to dygraph charts - that's because gauge and dygraph charts share the same chartIds and there's conflict (like the one with aggrMethod). Do you agree with that? The selected dimensions will still be passed to dygraphs, [here](https://github.com/netdata/dashboard/blob/7d6e115f5204a50d3ba5cc6152cc06a24c96521e/src/domains/dashboard/components/node-view/render-submenu-name.tsx#L121-L155)

VLegakis: Yeap I see! Nice work @jacekkolasa ! Thanks for the clarifications.

VLegakis (approved):

PR #153, opened 2020-10-15
Fix null response crash

VLegakis:


jacekkolasa (changes requested):
I think that will create an unstable state - dashboard will still see the call as "pending" and will not make any more requests (but i haven't checked)

Btw. are we accepting null results? It was never a problem before with old netdata, maybe BE should do the fix?

VLegakis:
> I think that will create an unstable state - dashboard will still see the call as "pending" and will not make any more requests (but i haven't checked)
> 
> Btw. are we accepting null results? It was never a problem before with old netdata, maybe BE should do the fix?

Thanks for that. So my initial approach was:
```
if (!data?.results) {
    fetchDataResponseChannel.put(fetchDataAction.failure({ id }))
} else {
...
```
Didn't know which one you prefer. Would you like that varsion resolving the saga with fail?

VLegakis:
O provided that initial version resolving the saga as failed. Please take a look and let me know @jacekkolasa 

(Please note  I think sagas don't have an issue remaining unresolved (they are cleaned up from saga framework as far as I have seen in the past on the Cloud) In addition to that on the cloud we use a generator paradigm so not sure about that here. 
You are right better play safe and always resolve them..

jacekkolasa:
that second approach should be fine! But i think you forgot to remove that previous solution

jacekkolasa (approved):
looks good! Disregard my latest comment about previous version, my mistake

PR #157, opened 2020-10-21
dashboard styling fixes

jacekkolasa:
- make header colors constant, because it's not yet ready for changing theme
- sign-in buttons hover fixes
- propagate Dashboard's theme from localStorage to styled-components
- social media links styling (fixed white theme, correct typo)
- hide small white dot, left from old iframe
- add more space on top so first header is visible below the new time-frame bar

### DIFF
--- a/src/components/app-header/styled.ts
+++ b/src/components/app-header/styled.ts
@@ -12,7 +12,7 @@ export const StyledHeader = styled.header`
   position: fixed;
   height: 56px;
   width: 100%;
-  background: ${getColor("tooltip")};
+  background: ${getColor(["gray", "limedSpruce"])};
   display: flex;
   flex-flow: row nowrap;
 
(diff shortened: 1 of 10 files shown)


src/components/app-header/styled.ts
```diff
@@ -97,19 +97,20 @@ export const SignInButton = styled.a<{ isDisabled: boolean }>`
...
-  color: #FFF;
   flex-flow: row nowrap;
   align-items: center;
-  text-decoration: none;
   user-select: none;
   display: flex;
   text-transform: uppercase;
   justify-content: center;
+  &, &:hover {
+    color: #FFF;
```
novykh, line 107: I guess we can use the bright theme color here

novykh (approved):

src/components/app-header/styled.ts
jacekkolasa: done

PR #158, opened 2020-10-26
Added group by attribute on charts.

novykh:


src/domains/chart/sagas.ts
```diff
@@ -136,9 +137,14 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
       group,
       gtime,
       agent_options: options.split("|"),
-      aggregation: {
-        method: aggrMethod,
+      aggregations: [groupBy === "node" && {
```
VLegakis, line 140: Not sure I understand this. Can you please explain what you have agreed for that with @lvrach ?
In case of `nodes` grouping I see this will be a list of aggregations? 
```
     [{
       method: "sum",
        groupBy: ["chart", "node"],
      },
      {
        method: aggrMethod,
        groupBy: [groupBy],
      }]
```
Can you explain this a bit?

VLegakis (approved):
Seem ok. Just wanted some clarifications if possible on what has been agreed

jacekkolasa (approved):
looks good! One thing however - the `groupBy` naming can be confusing for agent-dashboard use-case, since one could mistake it with `method` param (grouping across time in single node). Perhaps in future we should have more explicit naming for this

PR #160, opened 2020-10-26
submenu-info html fix

jacekkolasa:
related issue: https://github.com/netdata/netdata/issues/10118
Use dangerouslySetInnerHTML to display submenu info written in HTML in dashboard_info.js, just like we do it for other parts in node-view.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.7.2",
+  "version": "2.7.3",
   "lockfileVersion": 1,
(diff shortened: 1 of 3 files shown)


VLegakis (approved):
Looks nice @jacekkolasa !
What exactly does the change target? (I mean the case you have in mind for this fix)

jacekkolasa:
@VLegakis it's targeting the cloud's node-view issue with dashboard_info.js, as you can see in https://github.com/netdata/netdata/issues/10118. Probably it's also present in overview screen.

PR #164, opened 2020-10-29
Updated charts colors based on cloud theme.

novykh:
Fixed colors for charts based on [this](https://github.com/netdata/product/issues/1506) issue

### DIFF
--- a/src/domains/dashboard/components/node-view/use-update-theme.ts
+++ b/src/domains/dashboard/components/node-view/use-update-theme.ts
@@ -1,0 +1,64 @@
+import { useContext, useMemo } from "react"
+import { ThemeContext } from "styled-components"
+
+export const useUpdateTheme = () => {
+  const theme = useContext(ThemeContext)
+  const currentTheme = window.NETDATA.themes.current
+
+  useMemo(() => {
+    if (!theme.colors) return
+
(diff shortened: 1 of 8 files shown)


src/domains/dashboard/components/node-view/use-update-theme.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    window.NETDATA.themes.current = {
+      ...currentTheme,
+      background: theme.colors.elementBackground,
+      foreground: theme.colors.border,
+      grid: theme.colors.disabled,
+      axis: theme.colors.elementBackground,
+      highlight: theme.colors.attention,
+      colors: [
+        "#B0E952",
```
VLegakis, line 19: I suppose these match the theme colours used in the cloud. So couldn't we reference them from UIKit same way the cloud does? (`getColor("color-name...")` etc)?

novykh: Those are new actually, just for the charts. We can completely update the charts colors in the ui kit and use them here directly in the future.

VLegakis (approved):
Nice! Lets keep a note to reference these hard coded colours from the UIKit at some point

PR #171, opened 2020-11-05
Fix/update gauges and pie chart colors based on theme

allelos:
Update colors for gauges and pie charts based on [issue](https://github.com/netdata/product/issues/1506) issue

**Note**
I suppose that now we can remove the `colors` key from `useUpdateTheme` hook [here](https://github.com/netdata/dashboard/blob/master/src/domains/dashboard/components/node-view/use-update-theme.ts#L18-L41) since it's the same for all cases. I don't have enough context, so I'm just mentioning it here :)


apostolidhs (approved):

dashboard-react.js
```diff
@@ -249,77 +249,112 @@ NETDATA.statistics = {
...
+    ],
+    easypiechart_track: "#f0f0f0",
+    easypiechart_scale: "#dfe0e0",
+    gauge_pointer: "#93A3B0",
+    gauge_stroke: "#ECEFF2",
     gauge_gradient: false,
-    gauge_stop_color: '#FC8D5E',
-    gauge_start_color: '#B0E952',
+    gauge_stop_color: "#FF4C00",
+    gauge_start_color: "#FF4C00",
```
novykh, line 289: Please verify that this is not a mistake here. It seems weird to use the same color.

src/styles/dashboard.slate.css
```diff
@@ -1,5 +1,6 @@
 /* SPDX-License-Identifier: GPL-3.0-or-later */
 :root {
+    --color-main: #FFFFFF;
```
novykh, line 3: I'm sorry about this, but the `color-[primary|border]` are not used in here, we need to add them in the styles. You can check the light dashboard css for the lines affected.

novykh (approved):
You are right, let's remove the use-update-theme completely. Please check the comments if something needs to be addressed.

src/styles/dashboard.slate.css
allelos: Can you elaborate a bit more on this one? I am probably missing something

dashboard-react.js
allelos: We might have to get some input from Andrzej on this one! The design doesn't include a gradient, so I took it for granted that it should be that way

image

src/styles/dashboard.slate.css
allelos: I am using them in [main.css](https://github.com/netdata/dashboard/blob/15aca8895c7e699c07d94032347aac27ae145901/src/styles/main.css#L95-L103) so that's why I put them here so main.css can access them depending theme

novykh (approved):

dashboard-react.js
allelos: @novykh Andrzej just verified it. Flat colors only he said. So I suppose we go with this...

jacekkolasa (approved):
works good! I tested it also with /old dashboard and tv.html (they use dashboard.css and dashboard-slate.css files, too)

allelos:
> works good! I tested it also with /old dashboard and tv.html (they use dashboard.css and dashboard-slate.css files, too)

Thanks for the testing @jacekkolasa 🎸

PR #173, opened 2020-11-10
Group by node

apostolidhs:
- Adds the double aggregation logic that is required if the group is by node.
- Change chart type if it is required by the business logic

### DIFF
--- a/src/domains/chart/components/lib-charts/dygraph/utils.ts
+++ b/src/domains/chart/components/lib-charts/dygraph/utils.ts
@@ -37,6 +37,10 @@ export const getDataForFakeStacked = (
   ]
 })
 
+const isPercentage = (unit: string) => unit === "percentage"
+  || unit === "percent"
+  || unit.indexOf("%") !== -1
+
 export const getDygraphChartType = (
   attributes: Attributes, chartData: DygraphData, chartMetadata: ChartMetadata,
   chartSettings: ChartLibraryConfig,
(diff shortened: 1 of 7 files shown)


src/domains/chart/components/lib-charts/dygraph/utils.ts
```diff
@@ -37,14 +37,24 @@ export const getDataForFakeStacked = (
   ]
 })
 
+const isPercentage = (unit: string) => unit === "percentage"
```
novykh, line 40: `/(percent|%)/.test(unit)`

src/domains/chart/sagas.ts
```diff
@@ -137,17 +137,17 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
...
       agent_options: options.split("|"),
-      aggregation: {
-        method: aggrMethod,
-      },
-      // aggregations: [groupBy === "node" && {
-      //   method: "sum",
-      //   groupBy: ["chart", "node"],
-      // },
-      // {
+      // aggregation: {
```
novykh, line 140: Let's remove it all together.

novykh (approved):

src/domains/chart/components/lib-charts/dygraph/utils.ts
apostolidhs: I am trying to be as close to the business requirement as possible. I prefer exactly the word in order to avoid cases like 'percentile' or 'x/percent'

PR #176, opened 2020-11-12
Resized the chart when dimensions change.

apostolidhs:
Resize the chart when the names in the legend are getting changed. So we avoid the issue in the following schreenshot.

Screenshot 2020-11-11 at 6 23 06 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.8.3",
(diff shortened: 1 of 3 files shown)


VLegakis (approved):

jacekkolasa (commented):
@apostolidhs nice solution! As a side note - that effect needs some fix - it triggers additional update after mount, but your change doesn't affect that (it's already happening because of `resizeHeight` dependency). good stuff

PR #179, opened 2020-11-14
Updated node-view to accept default chart as a prop.

novykh:
If prop defaultChart exists, then view should scroll to this chart element.

### DIFF
--- a/src/domains/dashboard/components/node-view/node-view.tsx
+++ b/src/domains/dashboard/components/node-view/node-view.tsx
@@ -163,7 +163,7 @@ export const NodeView = ({
     const timeoutID = setTimeout(() => {
       const chartElement = document.querySelector(`#${defaultChart}`)
       if (chartElement) chartElement.scrollIntoView()
-    }, 100)
+    })
     // eslint-disable-next-line consistent-return
     return () => clearTimeout(timeoutID)
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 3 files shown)


src/domains/dashboard/components/node-view/node-view.tsx
```diff
@@ -144,16 +145,27 @@ export const NodeView = ({
...
 
+  useEffect(() => {
+    if (!defaultChart) return
+
+    setTimeout(() => {
+      const chartElement = document.querySelector(`#${defaultChart}`)
+      if (chartElement) chartElement.scrollIntoView()
+    }, 100)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
apostolidhs, line 168: It would be better to clear the timeout id on unmount.
Are we sure that there is not any safer way to ensure that the charts have been mounted? Maybe an `onReady` callback?

apostolidhs (approved):

VLegakis:
Seems ok with me. However I'm pretty sure this functionality was achieved through hash params. I'd leave @jacekkolasa comment-review on this

src/domains/dashboard/components/node-view/node-view.tsx
jacekkolasa: does `0` works the same as `100`? `0` is still a timeout-hack, but at least it's safer against race-conditions :)

jacekkolasa (approved):

PR #190, opened 2020-11-24
Minor perf update - debouncing scrolling affected useEffect

novykh:


jacekkolasa (approved):
good idea! Maybe we can improve it in the future by having the scroll-listener debounced, to avoid `useScroll` updating the component, but probably due to memoizations it won't be needed

novykh:
> good idea! Maybe we can improve it in the future by having the scroll-listener debounced, to avoid `useScroll` updating the component, but probably due to memoizations it won't be needed

Actually it does rerender since it's using state internally. So, yes I agree with the proposal to remove the hook. I can do it on a next PR.

PR #199, opened 2020-12-08
fix flip when BE doesn't support the lack of this flag

jacekkolasa:
related to https://github.com/netdata/product/issues/1469

### DIFF
--- a/src/domains/chart/sagas.ts
+++ b/src/domains/chart/sagas.ts
@@ -85,6 +85,9 @@ const constructCompatibleKey = (dimensions: undefined | string, options: string)
   },${encodeURIComponent(options)}`
 )
 
+// currently BE always transforms data as if `flip` was there
+const IS_FLIP_RESPECTED_IN_COMPOSITE_CHARTS = false
+
 const [fetchMetrics$] = getFetchStream(
   isPrintMode ? CONCURRENT_CALLS_LIMIT_PRINT : CONCURRENT_CALLS_LIMIT_METRICS,
 )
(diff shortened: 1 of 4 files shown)


src/domains/chart/sagas.ts
```diff
@@ -123,6 +126,13 @@ function* fetchDataSaga({ payload }: Action<FetchDataPayload>) {
     ? `${alwaysEndWithSlash(host)}api/v1/data`
     : host
 
+  const agentOptionsOriginal = options.split("|")
+  const hasFlip = agentOptionsOriginal.includes("flip")
+  const shouldAddFakeFlip = !IS_FLIP_RESPECTED_IN_COMPOSITE_CHARTS && !hasFlip
+  // if flip is not respected in composite-charts, send it always (like dygraph charts normally do)
+  const agentOptions = shouldAddFakeFlip
+    ? agentOptionsOriginal.concat("flip") : agentOptionsOriginal
+
```
apostolidhs, line 135: Since the `split, includes` are executed once the flag `IS_FLIP_RESPECTED_IN_COMPOSITE_CHARTS` is true, i suggest change the code in order to execute the extra code only when it is required. Maybe something like `transformOprtions(options) {if (IS_FLIP_RESPECTED_IN_COMPOSITE_CHARTS) return options.split("|") else ...}`

apostolidhs (approved):
Can't we wait for the backend to apply this change?

src/domains/chart/sagas.ts
jacekkolasa: `split` needs to be done in all conditions actually. So it's just the `includes` check, don't you think it's over-optimisation since this is code is only running when doing /data calls - and `includes` check takes only fraction of a ms?
My intention was to have the flag calculated once so it can be later used when transforming the response.

jacekkolasa:
> Can't we wait for the backend to apply this change?

@apostolidhs I'm not sure :) When i did it there was no fix from BE side, and this is IMO highest priority since we show wrong metrics. I'll check the progress on BE and we can decide then if that should be released.

jacekkolasa:
merging since approved, but we should be able to revert that later anyway

PR #203, opened 2020-12-16
dygraph panning responsiveness

jacekkolasa:
I've changed the way we update dygraph during panning.

- we always keep "requested" timeframe instead of timeframe that we get from data - so consecutive requests don't trigger when request for "similar" timeframe is already pending
- similar to the old dashboard, when user pans we debounce the requests until user pauses for 400 ms or there's a mouse-up event
- i've added a separate dependency - "useDebouncedCallback" from "use-debounce" hook that works much better than the one from "react-use" library, because it doesn't introduce a separate state, instead it works only on callbacks
- improved few memoizations, especially `dimensions` and `colours`, preventing some unnecessary updates

### DIFF
--- a/src/domains/chart/components/chart.tsx
+++ b/src/domains/chart/components/chart.tsx
@@ -3,6 +3,7 @@ import React, {
   useEffect, useState, useCallback, useMemo, memo, useContext,
 } from "react"
 import { ThemeContext } from "styled-components"
+import { useDebouncedCallback } from "use-debounce"
 
 import {
   requestCommonColorsAction,
@@ -38,6 +39,13 @@ import { LegendToolbox } from "./legend-toolbox"
 import { ResizeHandler } from "./resize-handler"
 import { AbstractChart } from "./abstract-chart"
 
+interface GlobalPanAndZoomState {
+  after: number // timestamp in ms
+  before: number // timestamp in ms
+  masterID?: string
+  shouldForceTimeRange?: boolean
+}
+
 interface Props {
   attributes: Attributes
   chartContainerElement: HTMLElement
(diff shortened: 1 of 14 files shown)


src/domains/chart/components/chart.tsx
```diff
@@ -121,7 +127,9 @@ export const Chart = memo(({
       )
       return dimensionNamesFromMetadata.concat(additionalDimensionNamesFromData)
     },
-    [chartData.dimension_names, chartMetadata.dimensions],
+    // intentionally skipping chartData.dimension_names for better memoization
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [dimensionNamesJoined, chartMetadata.dimensions],
```
apostolidhs, line 132: I think that this change adds overhead in the rerender lifecycle, since the concatenation of a large array is a heavy operation for both memory and processing utilisation. I suggest 2 alternatives:

* Shallow comparison of the `dimension_names`. So, `useShallowCompareEffect()`. This is also not optimal, but it is better that the `.join()`
* In the reducer that handles the response of the `chartData`, compare the current with the previous `dimension_names` and do not update this segment of the object.

apostolidhs (approved):

src/domains/chart/components/chart.tsx
jacekkolasa: @apostolidhs I like both options, and especially the second one :) I've fixed the reducer and added tests to prevent potential regressions in the future.

As a side note, I don't understand what is the performance benefit, because I think .join() takes a very little fraction of a `ms`, and we don't display more than 20 charts at the same time. So this is IMHO unnoticeable difference (to compare, this PR prevented additional dygraph rerender which takes 20-30 ms + DOM update) and code readability is what helps us preventing really expensive updates. But both alternatives that you provided seem to be cleaner than a hack with creating separate hash, so I completely agree with this change :)

apostolidhs: Sorry, i did not describe it properly. I was meaning that this change (the `.join()`) is much better than the previous implementation, but since we are making a change there we could have a better approach.

I think that your implementation (in reducer) is the best approach :)

PR #205, opened 2020-12-18
Supported multiple kubernetes cluster ids.

apostolidhs:
So we can support collections like 

Screenshot 2020-12-18 at 7 59 03 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.12.0",
+  "version": "2.12.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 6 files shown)


src/domains/chart/chart-types.ts
```diff
@@ -70,7 +70,7 @@ export interface ChartMetadata {
   dimensions: {
     [key: string]: Dimension
   }
-  chartLabels?: {
```
jacekkolasa, line 73: We don't always have that in `ChartMetadata` payload.

You can instead add question mark back, and in `render-charts-and-menu` use `!`, for example: `chartEnriched.chartLabels!.k8s_cluster_id` by which you signalise that this property is not `undefined`, neither `null`.

allelos (approved):

PR #208, opened 2021-01-05
fix dygraph's iframes overlapping

jacekkolasa:
fix problematic dygraph's feature, to limit "overlapping" of third-party iframes, which causes unwanted scrollbar on panning

image


jacekkolasa:
thx @allelos , btw. FYI i haven't disabled that functionality completely because we actually use it, we have 3 iframes in Dashboard :)

PR #209, opened 2021-01-05
Documentation helper

apostolidhs:
Replaced documentation view with https://github.com/netdata/netdata-ui/pull/168


allelos (approved):

src/components/app-header/app-header.tsx
```diff
@@ -199,16 +199,19 @@ export const AppHeader = ({
...
-          />
+          <Documentation app="agent">
+            {(toggle: any) => (
+              <Button
+                flavour="borderless"
+                neutral
+                themeType="dark"
+                className="btn"
+                icon="question"
+                onClick={toggle}
```
jacekkolasa, line 210: what do you think about using `onToggle` or `handleToggle` naming convention?
Especially since we'll remove typescript, IMO it would be good to have some hint what is type is that
cc @allelos @novykh 
https://medium.com/javascript-in-plain-english/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

public/index.html
```diff
@@ -444,119 +444,6 @@ <h4 class="modal-title" id="welcomeModalLabel">Welcome to the world of netdata</
         </div>
     </div>
 
-    <div class="modal fade" id="helpModal" tabindex="-1" role="dialog" aria-labelledby="helpModalLabel">
```
jacekkolasa, line 447: +1

jacekkolasa (approved):

src/components/app-header/app-header.tsx
apostolidhs: I think that it depends on the case. For instance, the `toggle` verb cannot be used as a standalone variable like `loading, loaded` etc., it describes a reaction, so the `onToggle` seems redundant to me. 

We can take this example
https://github.com/streamich/react-use/blob/master/docs/useToggle.md
```
const [on, toggle] = useToggle(true);
..onClick={toggle}..
```

Since it is not something major, we could also wait for the other's opinion.

PR #226, opened 2021-02-09
Kubernetes data testid

apostolidhs:
* Make node clickable
* Added testids
* UX improvements

### DIFF
--- a/src/domains/chart/components/lib-charts/group-box-chart/kubernetes/chartOverview.tsx
+++ b/src/domains/chart/components/lib-charts/group-box-chart/kubernetes/chartOverview.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable object-curly-newline */
 /* eslint-disable react/jsx-props-no-spreading */
 /* eslint-disable react/destructuring-assignment */
 /* eslint-disable operator-linebreak */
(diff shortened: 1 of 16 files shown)


src/domains/chart/components/lib-charts/group-box-chart/kubernetes/chartOverview.tsx
```diff
@@ -39,8 +48,24 @@ const ChartValueContainer = memo(({ id, units, displayedIndex }) => {
...
+    <Text
+      wordBreak="keep-all"
+      color={["white", "pure"]}
+      margin={[0, 0, 0, "auto"]}
+      data-testid="k8sPopoverChart-chartValue"
+    >
+      {aggregation && (
+        <Text
+          margin={[0, 1, 0, 0]}
+          color={["gray", "nepal"]}
```
novykh, line 63: none of the semantic colors is working here? Make sure it's ok on theme change.

src/domains/chart/components/lib-charts/group-box-chart/kubernetes/header.tsx
```diff
@@ -4,6 +4,8 @@
 import React from "react"
 import { H5 } from "@netdata/netdata-ui"
 
-const Header = (props) => <H5 color={["white", "pure"]} wordBreak="break-all" {...props} />
+const Header = (props) => (
+  <H5 color={["white", "pure"]} wordBreak="break-all" data-testid="k8sPopover-header" {...props} />
```
novykh, line 8: Same, maybe use `bright`?

jacekkolasa (approved):

novykh (approved):

src/domains/chart/components/lib-charts/group-box-chart/kubernetes/header.tsx
apostolidhs: The problem is that the popover background color is always dark, so all the typography component should be always white.

PR #232, opened 2021-02-16
Charts fixes

apostolidhs:
General fixies around the chart styling and functionality.

### DIFF
--- a/src/domains/chart/hooks/use-proceeded-chart.ts
+++ b/src/domains/chart/hooks/use-proceeded-chart.ts
@@ -1,21 +1,27 @@
+/* eslint-disable function-paren-newline */
+/* eslint-disable comma-dangle */
 import React, { useRef } from "react"
 import { useToggle } from "react-use"
 
-const useProceededChart = (firstEntry: number):
-  [boolean, React.Ref<HTMLElement>, (g: Dygraph) => void] => {
+const useProceededChart = (
+  chartRef: any,
+  propsRef: any
+): [boolean, React.Ref<HTMLElement>, (g: Dygraph) => void] => {
   const [proceeded, toggleProceeded] = useToggle(false)
 
   const ref = useRef<HTMLElement>(null)
 
   const updatePosition = (g: Dygraph) => {
     const { x } = g.getArea()
-    const distance = g.toDomXCoord(firstEntry * 1000)
+    const distance = g.toDomXCoord(propsRef.current.chartData.first_entry * 1000)
     const hasProceeded = distance > x
     toggleProceeded(hasProceeded)
 
     if (hasProceeded && ref.current) {
+      const { height } = chartRef.current.getBoundingClientRect()
       ref.current.style.left = `${x}px`
       ref.current.style.right = `calc(100% - ${distance}px)`
+      ref.current.style.top = `${height / 2}px`
     }
   }
 
--- a/src/domains/chart/hooks/use-proceeded-chart.ts
+++ b/src/domains/chart/hooks/use-proceeded-chart.ts
@@ -1,6 +1,7 @@
+/* eslint-disable react-hooks/exhaustive-deps */
 /* eslint-disable function-paren-newline */
 /* eslint-disable comma-dangle */
-import React, { useRef } from "react"
+import React, { useRef, useCallback } from "react"
 import { useToggle } from "react-use"
 
 const useProceededChart = (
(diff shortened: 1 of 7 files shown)


src/styles/dashboard.css
```diff
@@ -216,14 +216,15 @@ body {
 
 .netdata-chart-legend {
     position: absolute; /* within .netdata-container */
-    top: 0;
+    top: 26px;
+    bottom: 18px;
```
allelos, line 220: 👀

allelos (approved):

src/domains/chart/hooks/use-proceeded-chart.ts
```diff
@@ -1,21 +1,27 @@
...
+  propsRef: any
+): [boolean, React.Ref<HTMLElement>, (g: Dygraph) => void] => {
   const [proceeded, toggleProceeded] = useToggle(false)
 
   const ref = useRef<HTMLElement>(null)
 
   const updatePosition = (g: Dygraph) => {
     const { x } = g.getArea()
-    const distance = g.toDomXCoord(firstEntry * 1000)
+    const distance = g.toDomXCoord(propsRef.current.chartData.first_entry * 1000)
```
jacekkolasa, line 16: i've done similar fix, other way around :)
https://github.com/netdata/dashboard/pull/224/files#diff-13d03a896e7fe5f69e999196f24f547a72d9087d7302ad7dc0847b73347c9473R524

jacekkolasa: in https://github.com/netdata/dashboard/pull/224

src/styles/dashboard.css
jacekkolasa: @apostolidhs i think all the changes should be added also to `dashboard.slate.css`. Also, have you checked how it works with different legend position?
Apart from that, this file is meant to be included also in "static/custom dashboards" (html files). We haven't released react-dashboard as dashboard.js yet, but when we'll do it, it's better not to cause breaking changes in this case.

src/domains/chart/hooks/use-proceeded-chart.ts
apostolidhs: So, how will we proceed with this? 

There are 2 differences. 
* I fixed the vertical alignment
* Pass only the reference of the value in the function and not the entire function as a reference.

Since we need this functionality for the upcoming release, do you want to rebase over this once i merge it and continue with that logic?

jacekkolasa: I haven't fixed the vertical alignment, so it's fine that you did it (i thought it was made that low on purpose, because charts often take the bottom part of the container, but centered looks better).

As for the ref - I think the root of the problem is passing callbacks to `new Dygraph()`, and those callbacks are becoming outdated. So i changed it this way because I wanted to fix the source of problem, not the side effect of it. That solution is also existing for other callbacks (`updateChartPanOrZoom` etc.), and probably should be added to few more (for example `setHoveredX`, which works now but it can be problematic in the future).

So if the callback function is replaced with new function, IMO we should always use the newest one. If you put the value in ref, you essentially are calling the old function, which stays in memory because of this. But your solution works too, so if you feel better about this one, we can have this, up to you :)

I can rebase after your merge, no problem.

apostolidhs: My opinion is to try to avoid passing entire functions as a reference since it seems an antipattern in terms of reusability and composition.

Normally, when there is a function reference change, we should unbind and bind the callback from the Dygraph instance. (this `useEffect` is an example https://github.com/netdata/dashboard/blob/3e4c12edddf88d7cb0a42b4b8c0e7a0cf0d835c8/src/domains/chart/components/lib-charts/sparkline-chart.tsx#L191 )

Since we do not do that in most cases, i would prefer using the reference values

jacekkolasa: Unfortunately most of the time there is no way to unbind callbacks in the dygraph instance. And in cases where it is possible, there is big performance impact (dygraph is still a bottleneck here).

How is passing function as ref an anti pattern when it comes to reusability? You update ref only once, in the exact place where the binding breaks (where function is replaced by the new one), and if you use "value ref" pattern, you need to be aware where you use that value and selectively wrap with refs for every usage. And that callback is sometimes 3 levels up the tree, so you would have to wrap value with ref in chart.tsx, just because in dygraph the callback is not refreshed. :)
In other words - `useProceededChart` now breaks isolation, because it needs to "know" about implementation detail of `dygraph-chart`

PR #231, opened 2021-02-16
Componentize node view

apostolidhs:
* Replaced relative paths in dashboard build.
* Transform the charts to the menu structure.
* The menu structure contains all the required sections (head charts, title, icons etc..), sorted ready to be displayed by the components.
* Created React contexts in order to store the menu structure.
* Created the sidebar using menu context and composition.

## Notes
* The alias should be in a different PR, sorry for this.
* In order to make the sidebar work properly in the current ecosystem there are some minor compromises like the href ids etc. It Will be fully replaced once we refactor the main node-view screen
* Take a look it is not completely finished yet. I want to exhausted test it on cloud's single node & room overview screen.

PR #242, opened 2021-02-26
Virtualize the dashboard view

apostolidhs:
Included the https://github.com/netdata/dashboard/pull/231 PR.


jacekkolasa (commented):
Doing fix in this place can have a nice impact :)
One suggestion - IMO it is worth to check if it makes sense to omit `DisableOutOfView` - which is hacky virtualization based on IntersectionObserver. It can be redundant to have both.

jacekkolasa (approved):

Issue #240, opened 2021-02-25
mask referrer attributes in posthog. 


andrewm4894:
We should have a list of "allowed_referrer_domains"

e.g. 

```
allowed_referrer_domains = ['$direct', 'www.google.com', 'duckduckgo.com', 'app.netdata.cloud', 'www.reddit.com', 'my-netdata.io', 'github.com', 'www.netdata.cloud', 'staging.netdata.cloud']
```

- If `$initial_referring_domain` is not in `allowed_referrer_domains` then we should hash the values of `$initial_referrer` and `$initial_referrer_domain` and set them in the posthog.register(). 
- If `$referring_domain` is not in `allowed_referrer_domains` then we should hash the values of `$referrer` and `$referrer_domain` and set them in the posthog.register(). 

This will avoid any data we don't want like ip addresses etc from coming in via those attributes.

andrewm4894:
added some small comments to the PR

Issue #241, opened 2021-02-26
unmask certain event attributes on demo sites


andrewm4894:
For demo sites we actually should not mask certain event attributes and understanding a bit more about our demo site traffic will actually be very useful for various things. 

So for demo sites we should just not register the masked values for the following attributes: $current_url, $pathname, $host 

Instead we should just let them be whatever the normal way posthog captures them are. 

I think we can do this by just not setting those attributes in the `posthog.register()` when on netdata demo sites.

jacekkolasa:
it should be available today on nightlies

PR #243, opened 2021-03-01
mask referrer in posthog

jacekkolasa:
- Change several posthog bindings
- remove GTM and enable posthog without feature-flag
- add separate bindings for demo servers

fixes https://github.com/netdata/dashboard/issues/240,
https://github.com/netdata/dashboard/issues/241
https://github.com/netdata/dashboard/issues/244
https://github.com/netdata/dashboard/issues/245
https://github.com/netdata/dashboard/issues/246
https://github.com/netdata/dashboard/issues/248

### DIFF
--- a/src/domains/global/sagas.ts
+++ b/src/domains/global/sagas.ts
@@ -111,9 +111,9 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
     filter((value) => value !== undefined && value !== null,
       {
         $ip: "127.0.0.1",
-        $current_url: "agent dashboard",
-        $pathname: "netdata-dashboard",
-        $host: "dashboard.netdata.io",
+        $current_url: isDemo ? null : "agent dashboard",
+        $pathname: isDemo ? null : "netdata-dashboard",
+        $host: isDemo ? null : "dashboard.netdata.io",
 
         $initial_referring_domain: shouldMaskReferrer ? "127.0.0.1" : null,
         $initial_referrer: shouldMaskReferrer ? "http://127.0.0.1" : null,
--- a/src/domains/global/sagas.ts
+++ b/src/domains/global/sagas.ts
@@ -28,6 +28,7 @@ import {
 import { alarmsSagas } from "./alarms-sagas"
 import { MASKED_DATA } from "./constants"
 import { selectFullInfoPayload } from "./selectors"
+import { isAllowedReferrer } from "./utils"
 import { InfoPayload } from "./__mocks__/info-mock"
 
 const windowFocusChannel = channel()
@@ -104,6 +105,7 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
       }
     },
   })
+  const shouldMaskReferrer = !isDemo && !isAllowedReferrer(document.referrer)
   window.posthog.register(
     // remove properties with unavailable values
     filter((value) => value !== undefined && value !== null,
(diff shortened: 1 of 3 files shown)


src/domains/global/sagas.ts
```diff
@@ -104,14 +105,23 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
...
       {
         $ip: "127.0.0.1",
-        $current_url: "agent dashboard",
-        $pathname: "netdata-dashboard",
-        $host: "dashboard.netdata.io",
+        $current_url: isDemo ? null : "agent dashboard",
+        $pathname: isDemo ? null : "netdata-dashboard",
+        $host: isDemo ? null : "dashboard.netdata.io",
+
+        $initial_referring_domain: shouldMaskReferrer ? "127.0.0.1" : null,
```
andrewm4894, line 118: Can we use "masked" instead of "127.0.0.1"

src/domains/global/sagas.ts
```diff
@@ -104,14 +105,23 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
...
+        $initial_referrer: shouldMaskReferrer ? "http://127.0.0.1" : null,
```
andrewm4894, line 119: Can we use "masked" instead of "http://127.0.0.1"

src/domains/global/sagas.ts
```diff
@@ -104,14 +105,23 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
...
+        $referring_domain: shouldMaskReferrer ? "127.0.0.1" : null,
```
andrewm4894, line 120: Can we use "masked" instead of "127.0.0.1"

src/domains/global/sagas.ts
```diff
@@ -104,14 +105,23 @@ function* injectPosthog(machineGuid: string, personGuid?: string) {
...
+        $referrer: shouldMaskReferrer ? "http://127.0.0.1" : null,
```
andrewm4894, line 121: Can we use "masked" instead of "http://127.0.0.1"

src/domains/global/utils.ts
```diff
@@ -0,0 +1,11 @@
...
+  "",
+  "https://www.google.com/",
+  "https://duckduckgo.com/",
+  "https://www.reddit.com/",
+]
+
+export const isAllowedReferrer = (referrer: string) => allowedReferrerDomains.includes(referrer)
+  || referrer.endsWith(".my-netdata.io/")
+  || referrer.startsWith("https://github.com/")
+  || referrer.endsWith("netdata.cloud/")
```
andrewm4894, line 11: Should we add a 
```
|| referrer.startsWith("https://app.netdata.cloud/")
```

As i do also see some referrer strings like this:

https://app.netdata.cloud/spaces/andrewm4894/rooms/general/overview

Which could be very useful.

andrewm4894 (changes requested):
Just one or two small changes as i want to be as explicit as possible about data we mask so think using "masked" as opposed to "127.0.0.1" should be clearer for anyone using or looking at the data downstream anywhere.

novykh (approved):

src/domains/global/sagas.ts
jacekkolasa: sure, done in all 4 lines that you mentioned

PR #252, opened 2021-03-12
alarm-overlays

jacekkolasa:
Display a alarm overlay and a badge when dashboard is accessed with specific hash params.
Expose parts of it so cloud-fe can reuse it.

This is how it looks like, but most likely we'll be changing it in following iterations

obraz

### DIFF
--- a/src/domains/chart/components/disable-out-of-view.tsx
+++ b/src/domains/chart/components/disable-out-of-view.tsx
@@ -6,7 +6,7 @@ import { forEachObjIndexed } from "ramda"
 
 import { useDispatch, useSelector } from "store/redux-separate-context"
 import { isPrintMode } from "domains/dashboard/utils/parse-url"
-import { selectDestroyOnHide, selectIsAsyncOnScroll } from "domains/global/selectors"
+import { selectDestroyOnHide, selectIsAsyncOnScroll, selectAlarm } from "domains/global/selectors"
 import { getPortalNodeStyles } from "domains/chart/utils/get-portal-node-styles"
 import { Attributes } from "domains/chart/utils/transformDataAttributes"
 import { chartLibrariesSettings } from "domains/chart/utils/chartLibrariesSettings"
@@ -65,12 +65,12 @@ export const DisableOutOfView = ({
   /* separate functionality - adding custom styles to portalNode */
   const chartSettings = chartLibrariesSettings[attributes.chartLibrary]
   const [hasPortalNodeBeenStyled, setHasPortalNodeBeenStyled] = useState<boolean>(false)
-  // todo omit this for Cloud
+  const shouldAddSpecialHeight = useSelector(selectAlarm)?.chart === attributes.id
   useLayoutEffect(() => {
     if (hasPortalNodeBeenStyled) {
       return
     }
-    const styles = getPortalNodeStyles(attributes, chartSettings)
+    const styles = getPortalNodeStyles(attributes, chartSettings, shouldAddSpecialHeight)
     forEachObjIndexed((value, styleName) => {
       if (value) {
         portalNode.style.setProperty(styleName, value)
@@ -79,7 +79,8 @@ export const DisableOutOfView = ({
     // eslint-disable-next-line no-param-reassign
     portalNode.className = chartSettings.containerClass(attributes)
     setHasPortalNodeBeenStyled(true)
-  }, [attributes, chartSettings, hasPortalNodeBeenStyled, portalNode, setHasPortalNodeBeenStyled])
+  }, [attributes, chartSettings, hasPortalNodeBeenStyled, portalNode, setHasPortalNodeBeenStyled,
+    shouldAddSpecialHeight])
   /* end of "adding custom styles to portalNode" */
 
 
--- a/src/domains/chart/components/disable-out-of-view.tsx
+++ b/src/domains/chart/components/disable-out-of-view.tsx
@@ -65,11 +65,14 @@ export const DisableOutOfView = ({
   /* separate functionality - adding custom styles to portalNode */
   const chartSettings = chartLibrariesSettings[attributes.chartLibrary]
   const [hasPortalNodeBeenStyled, setHasPortalNodeBeenStyled] = useState<boolean>(false)
-  const shouldAddSpecialHeight = useSelector(selectAlarm)?.chart === attributes.id
+  const isShowingAlarmOnChart = useSelector(selectAlarm)?.chart === attributes.id
   useLayoutEffect(() => {
     if (hasPortalNodeBeenStyled) {
       return
     }
+    const shouldAddSpecialHeight = isShowingAlarmOnChart
+      && attributes.chartLibrary === "dygraph"
+      && chartSettings.hasLegend(attributes)
     const styles = getPortalNodeStyles(attributes, chartSettings, shouldAddSpecialHeight)
     forEachObjIndexed((value, styleName) => {
       if (value) {
@@ -79,8 +82,8 @@ export const DisableOutOfView = ({
     // eslint-disable-next-line no-param-reassign
     portalNode.className = chartSettings.containerClass(attributes)
     setHasPortalNodeBeenStyled(true)
-  }, [attributes, chartSettings, hasPortalNodeBeenStyled, portalNode, setHasPortalNodeBeenStyled,
-    shouldAddSpecialHeight])
+  }, [attributes, chartSettings, hasPortalNodeBeenStyled, isShowingAlarmOnChart, portalNode,
+    setHasPortalNodeBeenStyled ])
   /* end of "adding custom styles to portalNode" */
 
 
--- a/src/domains/chart/components/disable-out-of-view.tsx
+++ b/src/domains/chart/components/disable-out-of-view.tsx
@@ -65,7 +65,7 @@ export const DisableOutOfView = ({
   /* separate functionality - adding custom styles to portalNode */
   const chartSettings = chartLibrariesSettings[attributes.chartLibrary]
   const [hasPortalNodeBeenStyled, setHasPortalNodeBeenStyled] = useState<boolean>(false)
-  const isShowingAlarmOnChart = useSelector(selectAlarm)?.chart === attributes.id
+  const isShowingAlarmOnChart = useSelector(selectAlarm)?.chartId === attributes.id
   useLayoutEffect(() => {
     if (hasPortalNodeBeenStyled) {
       return
--- a/src/domains/chart/components/lib-charts/alarmBadge.js
+++ b/src/domains/chart/components/lib-charts/alarmBadge.js
@@ -1,23 +1,26 @@
 import React, { forwardRef } from "react"
 import styled from "styled-components"
 
-export const getBackgroundColor = (status) => ({
+const backgroundColorMap = {
   WARNING: "#FFF8E1",
   CRITICAL: "#FFEBEF",
   CLEAR: "#E5F5E8",
-})[status] || null
+}
+export const getBackgroundColor = (status) => backgroundColorMap[status] || null
 
-export const getBorderColor = (status) => ({
+const borderColorMap = {
   WARNING: "#FFC300",
   CRITICAL: "#F59B9B",
   CLEAR: "#68C47D",
-})[status] || null
+}
+export const getBorderColor = (status) => borderColorMap[status] || null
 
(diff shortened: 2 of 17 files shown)


src/domains/chart/components/lib-charts/alarmBadge.js
```diff
@@ -0,0 +1,54 @@
+import React, { forwardRef } from "react"
+import styled from "styled-components"
+
+export const getBackgroundColor = (status) => ({
+  WARNING: "#FFF8E1",
+  CRITICAL: "#FFEBEF",
+  CLEAR: "#E5F5E8",
+})[status] || null
```
apostolidhs, line 8: Since most of the colors does not exist in the pallet can we give it some time and discuss with the designers about that? We could either reuse a color or add those colors in the pallet.

src/domains/chart/components/lib-charts/alarmBadge.js
```diff
@@ -0,0 +1,54 @@
...
+const Badge = styled.div`
+  display: inline-block;
+  border-radius: 36px;
+  padding: 2px 12px;
+  background: ${({ background }) => background};
+  border: 1px solid ${({ border }) => border};
+  color: ${({ color }) => color};
+  font-size: 12px;
+  font-weight: 700;
+`
```
apostolidhs, line 39: The Badge could be a composition of

`<Flex as={TextSmall} .../>`

src/domains/chart/components/lib-charts/alarmBadge.js
```diff
@@ -0,0 +1,54 @@
...
+
+export default forwardRef((
+  { alarm: { status, value }},
```
apostolidhs, line 42: This component could be more general and receive status and a label as properties.

src/domains/chart/components/lib-charts/alarmBadge.js
```diff
@@ -0,0 +1,54 @@
...
+  WARNING: "#FFF8E1",
+  CRITICAL: "#FFEBEF",
+  CLEAR: "#E5F5E8",
+})[status] || null
+
+export const getBorderColor = (status) => ({
+  WARNING: "#FFC300",
+  CRITICAL: "#F59B9B",
+  CLEAR: "#68C47D",
+})[status] || null
```
apostolidhs, line 14: This pattern is not so optimal since a new object is created in every execution. 
You could move out the hash map definition so it will be executed once on the initialization. Something like:

const borderColorMap = {warning: '...'}

getBorderColor = borderColorMap[status] || borderColorMap.clear

src/domains/chart/components/disable-out-of-view.tsx
```diff
@@ -65,12 +65,15 @@ export const DisableOutOfView = ({
...
+  const isShowingAlarmOnChart = useSelector(selectAlarm)?.chart === attributes.id
   useLayoutEffect(() => {
     if (hasPortalNodeBeenStyled) {
       return
     }
-    const styles = getPortalNodeStyles(attributes, chartSettings)
+    const shouldAddSpecialHeight = isShowingAlarmOnChart
+      && attributes.chartLibrary === "dygraph"
+      && chartSettings.hasLegend(attributes)
+    const styles = getPortalNodeStyles(attributes, chartSettings, shouldAddSpecialHeight)
```
apostolidhs, line 76: Which part of the alarm adds special height to the chart? As I see the alarm component is inside the boundaries of the chart component.
I am asking because I should update the virtualized dashboard as well.

src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -515,6 +526,25 @@ export const DygraphChart = ({
         underlayCallback(canvas: CanvasRenderingContext2D, area: DygraphArea, g: Dygraph) {
           updatePrecededPosition(g)
 
+          if (propsRef.current.alarm) {
+            const { alarm: currentAlarm } = propsRef.current
+
+            const alarmPosition = g.toDomXCoord(currentAlarm.when * 1000)
+            const fillColor = getBorderColor(currentAlarm.status)
+
+            if (fillColor) {
```
apostolidhs, line 535: I think that determining the alarm status visibility from the border color is a wrong pattern.
If we do not accept invalid statuses can we make this check when we add the alarm to the state, so we will never have an invalid alarm state?

src/domains/chart/hooks/useAlarmBadge.js
```diff
@@ -0,0 +1,30 @@
...
+      }
+      const { x } = g.getArea()
+
+      ref.current.style.left = `${x}px`
+      ref.current.style.right = `calc(100% - ${position}px)`
+      ref.current.style.top = "40px"
+    }
+  }
+
+  return [isRendered, ref, updatePosition]
```
apostolidhs, line 29: The `updatePosition` could be defined under a useCalllback.
 
Statements like
```
    if (!alarm) {
      if (isRendered) {
        toggleIsRendered(false)
      }
      return
    }
```

Could be simplified

```
if (!alarm) return toggleIsRendered(false)
```

```
      if (!isRendered) {
        toggleIsRendered(true)
      }
      toggleIsRendered(true) // if the state is already true it will not rerender.
```

src/domains/chart/hooks/useAlarmBadge.js
```diff
@@ -0,0 +1,30 @@
+import { useRef } from "react"
+import { useToggle } from "react-use"
+
+export default () => {
+  const [isRendered, toggleIsRendered] = useToggle(false)
+
+  const ref = useRef(null)
+
+  const updatePosition = (alarm, g, position) => {
+    if (!alarm) {
```
apostolidhs, line 10: If we rename the alarm parameter to something more general like `visible`.
Then the entire `useAlarmBadge` will be more general to `useBadge`

src/domains/chart/hooks/useAlarmBadge.js
```diff
@@ -0,0 +1,30 @@
...
+
+    if (ref.current) {
+      if (!isRendered) {
+        toggleIsRendered(true)
+      }
+      const { x } = g.getArea()
+
+      ref.current.style.left = `${x}px`
+      ref.current.style.right = `calc(100% - ${position}px)`
+      ref.current.style.top = "40px"
```
apostolidhs, line 25: Can we define what is the `40px`, or a simple comment?

src/domains/chart/utils/get-portal-node-styles.ts
```diff
@@ -80,6 +82,13 @@ export const getPortalNodeStyles: GetPortalNodeStyles = (
     height = heightFromLocalStorage.replace(/"/g, "")
   }
 
+  if (shouldAddSpecialHeight) {
+    const heightOverriden = isLegendOnBottom
+      ? window.innerHeight * 0.5
+      : window.innerHeight * 0.4
+    height = `${heightOverriden}px`
```
apostolidhs, line 89: Is this also working in the virtualized dashboard?

src/domains/dashboard/hooks/useAlarmFromUrl.js
```diff
@@ -0,0 +1,35 @@
...
+    if (alarmWhen) {
+      const alarmTime = Number(alarmWhen)
+
+      const alarmStatus = params["alarm_status"]
+      const alarmChart = params["alarm_chart"]
+      const alarmValue = params["alarm_value"]
+
+      dispatch(setAlarmAction({
+        alarm: {
+          chart: alarmChart,
```
apostolidhs, line 21: Since this is an chartId and not an entire chart, could we rename it to `chartId`?

src/domains/chart/components/disable-out-of-view.tsx
jacekkolasa: The idea is that when `alarm_when` hash param is in the url, then the chart should have fixed height to around 40% of the screen. And this adjustment happens only on mount - so it has similar effect as having `chart_height` properties in localStorage on init.

Correct me if i'm wrong, but if settings in localStorage didn't need any handling in virtualised dashboard, this shouldn't be a problem too

src/domains/chart/hooks/useAlarmBadge.js
jacekkolasa: > // if the state is already true it will not rerender.

Actually i think it will re-render, last time i checked `useState` doesn't handle opting-out from update, nor does react-use. (https://codesandbox.io/s/focused-sammet-brw2d is the official usetoggle demo if you want to experiment)


> The updatePosition could be defined under a useCalllback.

I didn't use `useCallback` because changing `updatePosition` doesn't trigger any re-renders

src/domains/chart/utils/get-portal-node-styles.ts
jacekkolasa: Yep, i checked it both in Agent Dashboard and Cloud's Node-View, it has impact only on initial mount, i put more info on the other comment

src/domains/chart/components/lib-charts/dygraph-chart.tsx
jacekkolasa: Very good point, i actually already encountered a bug caused by this :) , i'll do exactly what you describe

src/domains/chart/components/disable-out-of-view.tsx
apostolidhs: There is the `useChartEstimatedHeight` function used by the `src/domains/agents/components/node-view/dashboard/chartSubMenuIntersection.js` & `src/domains/rooms/components/room-overview/dashboard/chartSubMenuIntersection.js` in order to estimate the height. I don't know if it needs special handling additional to the `legend position/chartHeaderHeight` modifiers

src/domains/chart/hooks/useAlarmBadge.js
apostolidhs: This is probably related to the react version. In your example, if you change to React `16.12.0` it will not rerender.
The normal behavior for the `useState` is to prevent rerenders if we set the same value. My suggestion is to assume that it is working like this in order to write more readable code.

https://dev.to/sunflower/reactjs-if-it-is-setting-a-state-with-the-same-value-will-the-component-be-re-rendered-5g24

src/domains/chart/components/lib-charts/alarmBadge.js
jacekkolasa: sure, i agree, but for now i think we need to stay with those overrides. This is actually why i put those colours here... Because they are not solid and perhaps will change.

src/domains/chart/components/lib-charts/alarmBadge.js
jacekkolasa: Correct me if i'm wrong, but because we use explicit colors, we cannot pass them to `Flex`, right? ie. `border`, `color` and `background`?

src/domains/chart/components/lib-charts/alarmBadge.js
jacekkolasa: Done, although maybe we should talk about such cases, I would like to learn about your reasoning, i see it as micro- (or even nano-) optimisation, and was focusing only on code readability

src/domains/chart/components/disable-out-of-view.tsx
jacekkolasa: @apostolidhs I'm not sure either, but if it does - then perhaps you'll need (or we'll need) also to handle height from `localStorage.chart_height.{...}`? I think that handling localStorage will be much more common (here this is only about redirect from alarm notification). 

Also, in case you'll need that info - adding extra height based on `legendPosition` was for me just a temporary measure - in the future i suspect there will be need to also take dimensions.length into account. But i'm not 100% sure if it's possible at this time

src/domains/chart/hooks/useAlarmBadge.js
jacekkolasa: of course! :) Thanks, that's how i remembered it should work, I was confused when i did codesandbox checks recently :)
done

src/domains/chart/components/lib-charts/alarmBadge.js
apostolidhs: You are right but even if we are not using all the attributes, it is better to use as many as possible. For instance, 
the `font-size: 12px;` will be filled from the `TextSmall `, the `font-weight: 700;` will be filled by the `strong` attribute, etc.

apostolidhs (approved):

PR #256, opened 2021-03-29
hash manipulation fix

jacekkolasa:
Don't split hash params by white-space, to fix problems with rich alarm values

Also added other changes that already are in cloud-frontend, and exposed them so we can avoid duplication.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.15.1",
+  "version": "2.15.2",
   "homepage": ".",
   "main": "./lib/src/index-npm.js",
   "files": [
--- a/src/domains/chart/components/lib-charts/alarmBadge.js
+++ b/src/domains/chart/components/lib-charts/alarmBadge.js
@@ -28,6 +28,7 @@ const Container = styled.div`
   overflow: hidden;
   pointer-events: none;
   direction: rtl;
+  z-index: 10; // higher than chart
 `
 
 const Badge = styled.div`
--- a/src/domains/chart/components/lib-charts/alarmBadge.js
+++ b/src/domains/chart/components/lib-charts/alarmBadge.js
@@ -44,16 +44,18 @@ const Badge = styled.div`
 `
 
 export default forwardRef((
-  { status, label },
+  { isVisible, status, label },
   ref,
 ) => (
   <Container ref={ref}>
-    <Badge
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.14.2",
+  "version": "2.14.5",
```
novykh, line 3: Beware of new versions on master

src/domains/chart/components/lib-charts/alarmBadge.js
```diff
@@ -39,19 +40,22 @@ const Badge = styled.div`
...
 ) => (
   <Container ref={ref}>
-    <Badge
-      background={getBackgroundColor(status)}
-      border={getBorderColor(status)}
-      color={getColor(status)}
-    >
-      {label}
-    </Badge>
+    {isVisible && (
```
novykh, line 51: should we keep the container when not visible?

src/domains/chart/components/lib-charts/dygraph-chart.tsx
```diff
@@ -1132,7 +1135,7 @@ export const DygraphChart = ({
       )}
       {alarm && hasLegend && (
         // @ts-ignore
-        <AlarmBadge ref={alarmBadgeRef} status={alarm.status} label={alarm.value} />
+        <AlarmBadge isVisible={isBadgeVisible} ref={alarmBadgeRef} status={alarm.status} label={alarm.value} />
```
novykh, line 1138: If for the above comment, we don't need the container to be rendered when not visible, I prefer to handle the visibility here, rather passing a flag in the component.

novykh (approved):

src/domains/chart/components/lib-charts/dygraph-chart.tsx
jacekkolasa: Unfortunately we need to render Container in AlarmBadge, so we can use the ref to its DOM Node (in dygraph callback). If i had that ref unavailable at time of callback, i would need to keep params (applied to `ref.current.style`) in the state and i'm worried that it would make the logic behind it less clear.

I limited the nr of times `AlarmBadge` will need to render by having it under `alarm` conditional - it will only render when the current chart will have "show alarm mode". @novykh how does it sound?

Btw. probably this feature will be changed in next iteration

PR #259, opened 2021-04-02
Passed uuid chart wrapper.

novykh:


src/domains/dashboard/components/chart-wrapper.tsx
```diff
@@ -57,7 +57,7 @@ export const ChartWrapper = ({
         <ChartContainer
           attributes={attributes}
           chartMetadata={chartMetadata}
-          chartUuid={chartUuid}
+          chartUuid={`${id}-${uuid}-${chartUuid}`}
```
jacekkolasa, line 60: Just interested, is there any reason why `chartUuid` isn't enough?

jacekkolasa (approved):

src/domains/dashboard/components/chart-wrapper.tsx
```diff
@@ -43,7 +43,7 @@ export const ChartWrapper = ({
   //   ? { width: "100%", height: `${height}px` }
   //   : { width: "100%" }
 
-  const chartUuid = useMemo(() => uuid.v4(), [])
+  const chartUuid = useMemo(() => getUuid.v4(), [])
```
jacekkolasa, line 46: IMHO changing name like this doesn't bring value, if the former way is already used in other places, and it's remembered

Issue #269, opened 2021-04-30
Update release workflow to automatically trigger generation of a PR in the agent repo to update the dashboard version used there.


Ferroin:
This will trigger the workflow described in https://github.com/netdata/netdata/issues/11073, and will essentially fully automate the release process for the dashboard.

PR #270, opened 2021-05-03
Automatically trigger an agent PR on dashboard release.

Ferroin:
This will automatically run a workflow in the agent repo to generate a PR to update the agent to the newest version of the dashboard whenever a release that is not a prerelease is created.

This will be marked ready for review once netdata/netdata#11076 is merged, as it has a hard dependency on that PR.

Closes: #269 

I’ve tagged @netdata/cloud-fe for review to get a confirmation that the described behavior is what we want. The other review requests are to help ensure correctness of the change itself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,3 +30,12 @@ jobs:
           asset_name: dashboard.tar.gz
           asset_content_type: application/gzip
           upload_url: ${{ github.event.release.upload_url }}
+      - name: 'Trigger agent code update'
+        if: github.event.release.prerelease == false
+        uses: benc-uk/workflow-dispatch@v1
+        with:
+          token: ${{ secrets.NETDATABOT_GITHUB_TOKEN }}
+          repo: netdata/netdata
+          workflow: Dashboard Version PR
+          ref: refs/heads/master
+          inputs: {"dashboard_version": "${{ github.event.release.tag_name }}"}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,4 +38,4 @@ jobs:
           repo: netdata/netdata
           workflow: Dashboard Version PR
           ref: refs/heads/master
-          inputs: {"dashboard_version": "${{ github.event.release.tag_name }}"}
+          inputs: '{"dashboard_version": "${{ github.event.release.tag_name }}"}'


kaskavel (approved):

vkalintiris (approved):

Ferroin:
@netdata/cloud-fe I would appreciate if I could get some confirmation that the described behavior (trigger a dashboard update PR in the agent repo on any release that is not a prerelease) is consistent with how we want this to behave.

Ferroin:
> @Ferroin in my understanding, what we need is to `npm pack` or something dashboard latest from npm, when the agent gets released and merge it in the repo in the process.
> 
> Said that, this approach is kind of weird for 2 reasons:
> 1st of all, we need to get rid of this file all together since we need to stick with the npm release and not build an already built package again, just for this specific case (agent release that is). _but this is how it was since before, so ok_

The insistence on being tied to a GitHub release is not important, we just need the code built and available _somewhere_ for this approach to work. I’m just trying to keep the overall required changes to a minimum by reusing what we’re already doing. If so desired, we can just have the GHA workflow run `npm release` and have the agent side of things pull from NPM instead of the GitHub release. That would have to wait until we’re done with this whole series of changes though, because until we are bundling the (built/compiled/minified) dashboard code into the agent repo instead of downloading it on install such a cange would add a hard dependency on Node.JS for our users, which is something that we have been actively trying to avoid.

> 2nd, is there anyone who guarantees that the PR will go through 100% to the agent release that needs it? I see manual process everywhere  _run this process here, wait for it to finish, check the PR on a different repo and merge it_

The change in the agent repo needs to wait for whatever release build we do anyway, there is _no other option_. Doing things on the agent side as a PR also ensures we actually get an entry in the agent release notes for the dashboard (which, given how I’ve written the agent CI code that generates the PR, will also link back to the release notes for the dashboard), which is _really_ important for ensuring our users know what’s going on.

> 
> Finally this approach feels wrong (at least to me) because I'd expect something like this to happen on the netdata repo and not here - I mean the requester should handle the dependencies and not the other way around. Afaic netdata repo needs dashboard and should request it (from the npm) and not the other way around.

This change is just adding code to fire an event to inform the agent repo that there is a new version of the dashboard. All the actual tooling surrounding updating the data for it in the agent repo is in a workflow there. As long as we want the dashboard version locked in the agent and updated as soon as there is a new dashboard release, we have to have _something_ on the dashboard side of things that notifies the CI in the agent repo that there’s a new version.

> The solution I'd expect to see was on the agent release process:
> `npm pack dashboard (in needed location) -> git commit (or exit if no changes) -> git push`

This needs to _not_ happen as part of the agent release process. We need the new dashboard code as soon as it’s available so that the nightlies get it.

> 
> No PRs, no manual intervention, no weird coupling between repos.
> 
> But I can confirm your PR works, **if this is how we want to proceed** with this.

PR #275, opened 2021-05-12
Time controls pause play global

allelos:
Issue: [here](https://github.com/netdata/product/issues/1511)
Figma Designs: [here](https://www.figma.com/file/RgebwHKHvTCAlTmIxLsb6l/Cloud-and-Agent-navigation?node-id=2%3A10080)

**Global controls**
Playing
image
Paused
image
Force Play
image

**Play Options**
image


src/domains/global/reducer.ts
```diff
@@ -79,6 +84,7 @@ export type StateT = {
   }
   hoveredX: number | null
   hasWindowFocus: boolean
+  globalPause: boolean
```
jacekkolasa, line 87: I'm afraid that is a state duplication, but correct me if i'm wrong. Until now, the "pause" mode could be calculated based on globalPanAndZoom, hoveredX, etc. It could be done as a separate selector (something like computed property based on already existing state).
Now if you add one more property it suddenly becomes more complicated because you can have unstable/impossible states, you need to always keep both in sync.

@allelos What do you think?

allelos: It would be state duplication if `pause` was only dependent on `globalPanAndZoom`, `hoveredX`, etc. The `pause` state in that case would be a derived property. 
From now on pausing will not only react to the above dependencies but also to external dependencies such as of manually pausing. It describes a different state.
Also keep in mind that during a forced pause we would also like to have access to `globalPanAndZoom` and `hoveredX` properties for showing values when hovering, panning etc without messing with the paused state.
In my opinion, this would be a lot more complex without having a global play / pause state. Also, complexities of 1 to 1 relationships will be, most of the time, a lot simpler to use and reason about than derived ones etc.

jacekkolasa: @allelos can you briefly explain when will be the situation where `globalPause` will be `true`, but `globalPanAndZoom` and `hoveredX` will be `null`?

jacekkolasa: previously to manually pause, `setGlobalPanAndZoom` action was send.
My problem is that we should not change all occurences where `globalPanAndZoom` was used to manipulate play/pause, and switch to `globalPause`, do you agree?

Re: complexities, IMHO derived properties (vs state duplication) are just less buggy, because you need to handle less nr of different scenarios.

In this scenario, i would combine `globalPanAndZoom` with `globalPause`, for example by adding `wasManuallyTriggered` flag inside `globalPanAndZoom` object. But of course your way will work too, just we need to update always both properties - `globalPanAndZoom` and `globalPause`.

allelos: > @allelos can you briefly explain when will be the situation where `globalPause` will be `true`, but `globalPanAndZoom` and `hoveredX` will be `null`?

The simplest case would be when just clicking the play / pause button. 

> My problem is that we should now change all occurences where `globalPanAndZoom` was used to manipulate play/pause, and switch to `globalPause`, do you agree?

Yes and no. It's true that `globalPanAndZoom` and `globalPause` are closely related especially compared to `globalPause` and `hoveredX`. But they are not describing the same thing. `globalPause` is a superset of `globalPanAndZoom`. Pause happens because of pan and zoom, but  pan and zoom doesn't happen because of pausing. 

If we were to combine them I would go with `globalPause` containing pan and zoom instead of the opposite. We can do this in the future, but I don't see any problem (besides updating both) with having them decoupled , at least for now.

@jacekkolasa wdyt?

jacekkolasa: > The simplest case would be when just clicking the play / pause button.
I had situations like this (for example when people were clicking "pause" button below chart), in those cases setGlobalPanAndZoom was dispatched. So i was using different state for exactly the same action, that's why i was proposing that we unify this. Or maybe we don't need to since new refactor is coming anyway :)

src/utils/server-detection.ts
```diff
@@ -36,7 +36,7 @@ export const getPathFromScriptSource = (source: string) => {
 
 const getDefaultServer = () => {
   if (isDevelopmentEnv) {
-    return "http://localhost:19999"
+    return "http://n1.katsuna.com:19999/"
```
jacekkolasa, line 39: @allelos :)
```suggestion
    return "http://localhost:19999"
```

jacekkolasa (changes requested):
"Sync pan and zoom on all charts?" - "Don't Sync" option doesn't seem to work now under the new controls. If that option is set, the I am unable to play back the chart. Should we maybe remove that option then?

I'll "request changes" until we'll have answer because this seems important.
cc @manos-saratsis

allelos:
> "Sync pan and zoom on all charts?" - "Don't Sync" option doesn't seem to work now under the new controls. If that option is set, the I am unable to play back the chart. Should we maybe remove that option then?
> 
> I'll "request changes" until we'll have answer because this seems important.
> cc @manos-saratsis

Thanks @jacekkolasa. The thing is I can't seem to reproduce this issue. Can you elaborate more?

jacekkolasa (commented):
to reproduce:
- switch "Sync pan and zoom on all charts?" as "Don't sync"
- refresh page
- pan any chart to the past
- time toolbox changes state - it is now in "paused" mode for all charts, and it shouldn't - in "Don't sync" mode you should be able to pan individual charts without affecting others
- if you pan any chart, it is impossible to "Play" them back. You can only do it with double-click

My suggestion is to put back "Play" button when "Dont sync" mode is on, or remove that setting from options.

allelos:
Awesome, thanks @jacekkolasa. 
I still can't reproduce the play part. The play / pause button works for both cases `sync` and `Don't sync` for me.
But I do understand the opposite. I assume your worry is about being able to pan a single chart while othera are playing and don't global pause. 
According to product specs I think this is not supported / covered https://github.com/netdata/product/issues/1511 @manos-saratsis

manos-saratsis:
> Awesome, thanks @jacekkolasa.
> I still can't reproduce the play part. The play / pause button works for both cases `sync` and `Don't sync` for me.
> But I do understand the opposite. I assume your worry is about being able to pan a single chart while others are playing and don't global pause.
> According to product specs I think this is not supported / covered [netdata/product#1511](https://github.com/netdata/product/issues/1511) @manos-saratsis

Let's remove "Sync pan and zoom on all charts?“ from the Agent and always sync

jacekkolasa (approved):

src/utils/server-detection.ts
allelos: 😄

PR #281, opened 2021-06-28
Fix react warnings

novykh:


apostolidhs (approved):

src/domains/chart/components/lib-charts/gauge-chart.tsx
```diff
@@ -101,7 +101,9 @@ export const GaugeChart = ({
   )([min, max])
   // calling outside "useEffect" intentionally,
   // because it should update the values first, and only then render the chart in useEffect()
-  setMinMax([safeMin, safeMax])
+  useEffect(() => {
+    setMinMax([safeMin, safeMax])
+  }, [safeMin, safeMax])
```
jacekkolasa, line 106:
```
  // calling outside "useEffect" intentionally,
```
This comment above is now outdated :)

jacekkolasa (commented):
@novykh Please be aware that this change can make gauge and easy-pie-chart "flicker"/rerender when values formatting is changed after calling `setMinMax`

Before, we were calling `setMinMax` and then those values were transformed, and only then the chart was created.
Now, we'll first create the chart, and then update the chart once again with new formatting, which can cause visible effect on labels.

novykh:
@jacekkolasa could you please make a check to see if there are any visible issues? I can't understand if there is any diff, before and after.

novykh:
Also, regarding the change, it's invalid to update a local state of another component while rendering.

jacekkolasa:
@novykh It is invalid (deprecated), but it's still working for backwards compatibility. At the time i was writing this react wasn't complaining about this
I'll check later if i can replicate this issue. It should be visible when there are big values which are auto-transformed, for example millions of bytes, time units, etc.

PR #305, opened 2021-09-14
rename "NetData" to "Netdata"

andrewm4894:
Rename "NetData" to "Netdata" in a few old cases where it is still used.


allelos:
Closing this in because of #303.

PR #307, opened 2021-09-15
2.19.20

allelos:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.19.9",
+  "version": "2.19.20",
```
jacekkolasa, line 3: 20? :)

jacekkolasa (approved):

package.json
allelos: It's my lucky number 😆 I could swear the previous one was `2.19.19`

PR #312, opened 2021-09-22
Add CI to check CI workflows when they are changed.

Ferroin:
Yes, I know it’s kind of meta.

This adds static checking for the GitHub Actions workflows in this repository. The checks will only run when these files actually change (they will show as skipped otherwise) so there should be no significant impact on normal development.


Ferroin:
yamllint complaints are expected and benign.

PR #313, opened 2021-09-23
Removed datetimeformat polyfills

allelos:


apostolidhs (commented):
1. Are we sure that we don't want the polyfills. Are we fully compatible with the supported browsers?
2. What values are required from the default timezone? Could we set some default values instead of an empty object?

PR #314, opened 2021-09-28
rendering charts and menu: add anomaly_detection case

ilyam8:
Needed for https://github.com/vkalintiris/netdata/pull/8


jacekkolasa (approved):
awesome!

andrewm4894:
@jacekkolasa would this then need to go through a dashboard release and so to "get" it we would need to update the dashboard used by [our feature branch](https://github.com/vkalintiris/netdata/tree/anomaly-detection-mvp) (or update from master once netdata/netdata starts using the dashboard that has what we need).

PR #331, opened 2021-10-15
different approach for export-snapshot problem

jacekkolasa:
fix exporting snapshots when default relative `after` is changed (use`defaultAfter` instead of specified initially options.duretion)


jjtsou (approved):
Kudos for the quick fix!

PR #341, opened 2021-11-23
Fix heading and description styles

burbuli8ra:
This PR implements updates needed for issue [#2215](https://github.com/netdata/product/issues/2215)

* Remove icon from page heading
* Change HTML tag of page sub-headings
* Update color of page heading, page sub-heading and description
* Add scripts to update `cloud-frontend` on development mode

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.20.12",
+  "version": "2.20.13",
   "homepage": ".",
   "main": "./lib/src/index-npm.js",
   "files": [
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "@material/list": "^4.0.0",
     "@material/menu": "^2.3.0",
     "@material/menu-surface": "^1.0.0",
-    "@netdata/netdata-ui": "^1.6.33",
+    "@netdata/netdata-ui": "^1.7.18",
     "@rmwc/icon": "^5.7.2",
     "@rmwc/list": "^5.7.2",
     "@rmwc/menu": "^5.7.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@netdata/dashboard",
-  "version": "2.20.13",
+  "version": "2.20.14",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -119,6 +119,8 @@
     "clean": "./node_modules/rimraf/bin.js ./lib",
     "remove-unnecessary-dts": "rimraf ./lib/dashboard-react.d.ts",
     "ts-bundle": "npm run clean && tsc --noEmit false && node copy-assets && npm run remove-unnecessary-dts && npm run resolveAlias",
+    "cp-cloud": "cp -R ./lib ../cloud-frontend/node_modules/@netdata/dashboard/",
+    "to-cloud": "yarn ts-bundle && yarn cp-cloud",
```
novykh, line 123: We use npm on dashboard.

src/domains/charts/chartMenuGroup/title.js
```diff
@@ -7,8 +7,7 @@ export const ChartMenuGroupIcon = ({ icon, ...rest }) => (
 )
 
 export const ChartMenuGroupTitle = ({ icon, title, ...rest }) => (
-  <Flex as={H0} color="key" gap={2} {...rest}>
-    <ChartMenuGroupIcon icon={icon} />
+  <Flex as={H0} color="text" gap={2} {...rest}>
```
novykh, line 10: Let's remove the span below as well. Is it needed?

novykh (approved):

package.json
burbuli8ra: Updating it! Thanks! 🙏

src/domains/charts/chartMenuGroup/title.js
burbuli8ra: Not sure yet!
After syncing with design, I will proceed with removing it (and maybe use only an `H0` instead of `Flex` or I will revert the removal of icon...! 😄

burbuli8ra: @novykh after syncing with design, we are indeed removing the icon, only from `cloud-frontend` headings, so I have proceeded with the replacement of components

PR #363, opened 2022-02-14
Moved disk inodes and space under mount points.

novykh:
Fixes [#2896](https://github.com/netdata/cloud-frontend/issues/2896)

### DIFF
--- a/src/domains/charts/getChartMenu.js
+++ b/src/domains/charts/getChartMenu.js
@@ -6,7 +6,7 @@ export default (
   const subMenuId = family || "all"
   const clusterId = hasKubernetes && chartLabels?.k8s_cluster_id?.[0]
 
-  const [type] = id.split(".")
+  const [type, typeB] = id.split(".")
   const parts = type.split("_")
   const [part1, part2] = parts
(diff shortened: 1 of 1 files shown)


src/domains/charts/getChartMenu.js
```diff
@@ -57,6 +56,13 @@ export default (
     case "anomaly":
       return emit({})
 
+    case "disk":
+    case "mount": {
+      if (!part2) return emit({ menu: part1 })
+      if (part2 === "inodes" || part2 === "space") return emit({ menu: "mount" })
+      return emit({ menu: part1 })
```
burbuli8ra, line 63: Why not `if (part2 && ((part2 === "inodes" || part2 === "space"))) return emit({ menu: "mount" })`?
Just to avoid having the `return emit({ menu: part1 })` twice! 😄

jjtsou: We can just erase the whole ` if (!part2) return emit({ menu: part1 })` I think 🤔

burbuli8ra (approved):

PR #368, opened 2022-02-28
Add code coverage report

ivangergo3:
Added similar test-coverage report as we have in "cound-frontend"

So from now on, all tests will be executed when a pull request is being created.

And we will publish the results into google cloud, so we can check the current status [here](https://github.com/netdata/test-automation/wiki/Netdata-Test-Coverage-Report)

Could you please check this change? And if you have any concerns let me know


tsconfig.json
```diff
@@ -28,12 +24,6 @@
...
   },
-  "include": [
-    "./src/**/*"
-  ],
-  "exclude": [
-    "/node_modules/",
-    "**/*.test.ts",
-    "**/*.test.tsx"
-  ]
+  "include": ["./src/**/*", "src/vendor/dygraph-c91c859.min.js"],
```
dimko, line 27: I don't see this dygraph js included in the previous version of the tsconfig.json, what is the reason for this being added now?

.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
...
+    name: Run Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: "12.22.0"
```
dimko, line 24: @netdata/cloud-fe should the version be that one? It's the same for every PR

.github/workflows/npm-test.yml
```diff
@@ -1,35 +0,0 @@
...
-      - name: Checkout Sources
-        uses: actions/checkout@v2
-      - name: Prepare Node.js
-        uses: actions/setup-node@v1
-        with:
-          node-version: '12.x'
-      - name: Run `npm install`
-        run: npm install
-      - name: Run `npm run lint`
-        run: npm run lint
```
dimko, line 35: This file was deleted, but lint check wasn't moved to the other workflow

.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
+---
+name: tests
+
+on:
+  push:
+    branches: [master, develop]
```
novykh, line 6: Only master

.github/workflows/npm-test.yml
burbuli8ra: @ivangergo3 as @dimko maybe you should move the lint check to the new file!
If you check, the checks fail on lint! 😄

.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: "netdata-cloud-testing"
+          service_account_key: ${{ secrets.TEST_AUTOMATION_SERVICE_ACCOUNT }}
+          export_default_credentials: true
+
+      - name: Upload report to test-reports bucket
+        if: (steps.run-tests.outcome == 'failure' || steps.run-tests.outcome == 'success') && github.ref == 'refs/heads/develop'
+        run: |
+          gsutil -h "Cache-Control: max-age=0, no-store" cp /tmp/coverage_report gs://${{ secrets.TEST_AUTOMATION_STORAGE_BUCKET }}/${{ github.event.repository.name }}/coverage_report
```
burbuli8ra, line 54: Also the checks throw an error here for too line size!

.github/workflows/npm-test.yml
burbuli8ra: As @dimko mentioned, maybe the check for lint should be moved to the new file as well! 😄

.github/workflows/test.yml
```diff
@@ -0,0 +1,65 @@
+---
+name: tests
+
+on:
+  push:
+    branches: [master]
+  workflow_dispatch:
+    inputs:
```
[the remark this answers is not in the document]
ivangergo3: Good point, removed, update all pending workflows

.github/workflows/test.yml
ivangergo3: Updated in all pending workflows

burbuli8ra (approved):

novykh (approved):

ivangergo3:
@jjtsou, it seems like it was a bad idea to remove the original testing pipeline. Should I put it back or should we run the new with all pull requests?

ivangergo3:
@novykh , it seems like it was a bad idea to remove the original testing pipeline. Should I put it back or should we run the new with all pull requests?

novykh:
Run the new on pull_requests.

PR #377, opened 2022-03-16
tmp

jjtsou:


src/domains/charts/provider.js
```diff
@@ -41,4 +38,36 @@ const Provider = ({
...
+  activeSubMenuId,
+  chartIds,
+  getChart,
+  dashboardAttributes,
+  hasKubernetes,
+  children,
+}) => {
+  const { menuGroupIds, menuGroupById, subMenuById, menuChartsAttributeById } = useMemo(
+    () => getMenu(chartIds, getChart, { hasKubernetes }),
+    [chartIds, getChart]
```
libekonst, line 53: `hasKubernetes` seems to be missing from the dep array, is that intended?

jjtsou:
Closing this one as the solution was introduced on cloud-fe repo

PR #395, opened 2022-04-22
feat: prefer 'k8s_cluster_name' for Kubernetes menu if exists

ilyam8:
Fixes: netdata/product#2382. The k8s_cluster_name label was added in netdata/netdata#12638 (GKE only).


ilyam8:
Ahh, we don't have it as a chart label, it is a host label.

ilyam8:
Ok, this was a naive attempt. There is no way to access host_labels.

ilyam8:
This PR will work after netdata/netdata#12858 is merged.

amalkov:
FYI: @hugovalente-pm

novykh (approved):

PR #418, opened 2022-06-20
Different mem.available description for FreeBSD

MrZammler:
Adds a check for FreeBSD and uses a different description for `mem.available` chart on FreeBSD.

This change is already available in https://github.com/netdata/netdata/blob/222a3f59e82434033297560511595d80ada0886f/web/gui/dashboard_info.js#L1733-L1740, advised by @ilyam8 it needs to be here as well (thanks!).

Not sure how to test (don't want to break dashboard), this is just copy paste from the above change.

### DIFF
--- a/src/dashboard_info.js
+++ b/src/dashboard_info.js
@@ -1648,7 +1648,12 @@ netdataDashboard.context = {
     },
 
     'mem.available': {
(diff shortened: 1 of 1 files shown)


ilyam8:
@MrZammler I suggest updating everything (there are more changes), not just copying your changes.

PR #419, opened 2022-06-20
update dashboard_info.js

ilyam8:
Manual sync with https://github.com/netdata/netdata/blob/master/web/gui/dashboard_info.js

I did no tests (rendering infos on the Cloud), but this is a copy of the dashboard_info.sh file from the netdata repo. Should be no problems.


jacekkolasa (approved):
thanks @ilyam8

PR #420, opened 2022-06-20
Stable chart height fix

jacekkolasa:
fixes a problem when during scrolling, sometimes headers between charts were moving few pixels up and down, which was a sign of reflow

removed support of old dashboard localStorage keys, which could be problematic in rare cases (and that support is not needed anymore)

Also removed unused functionality of badges for metrics_correlation, which is not used by the cloud anymore


burbuli8ra (approved):
LGTM!

novykh (approved):

PR #439, opened 2022-08-17
dashboard info by domain

jacekkolasa:
related to https://github.com/netdata/netdata-cloud/issues/541

Build dashboard from dashboard_info.js, based on domain (room | node), for example:

```
  "disk.util": {
    room: {
      info: "Info visible only in room overview",
      heads: [netdataDashboard.gaugeChart("Utilization in room", "12%", "", "#FF5588")],
    },
    node: {
      info: "Info visible only in node-view",
    },

    colors: "#FF5588",
    heads: [netdataDashboard.gaugeChart("Utilization", "12%", "", "#FF5588")],
    info: "Disk Utilization measures the amount of time the disk was busy with something. This is not related to its performance. 100% means that the system always had an outstanding operation on the disk. Keep in mind that depending on the underlying technology of the disk, 100% here may or may not be an indication of congestion.",
  },
```

If `room` or `node` keys are present, properties available there override the config for given context.

### DIFF
--- a/src/dashboard_info.js
+++ b/src/dashboard_info.js
@@ -3907,10 +3907,10 @@ netdataDashboard.context = {
                 }
             ],
         },
-        info: '<p>A connection is an established line of communication between a client and the PostgreSQL server. Each connection adds to the load on the PostgreSQL server. To guard against running out of memory or overloading the database the <i>max_connections</i> parameter (default = 100) defines the maximum number of concurrent connections to the database server. A separate parameter, <i>superuser_reserved_connections</i> (default = 3), defines the quota for superuser connections (so that superusers can connect even if all other connection slots are blocked).</p><p></p><p><b>Total connection utilization across all databases.</b> Utilization is measured as a percentage of (<i>max_connections</i> - <i>superuser_reserved_connections</i>). If the utilization is 100% no more new connections will be accepted (superuser connections will stil be accepted if superuser quota is available).</p>'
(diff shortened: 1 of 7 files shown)


src/dashboard_info.js
```diff
@@ -3882,10 +3887,30 @@ netdataDashboard.context = {
...
     },
     'postgres.connections_usage': {
-        info: '<p>Connections usage. The maximum number of concurrent connections to the database server is <i>max_connections</i> minus <i>superuser_reserved_connections</i>.</p><p><b>Available</b> - new connections allowed. <b>Used</b> - connections currently in use.</p>'
+        info: '<p><b>Connections usage.</b> The maximum number of concurrent connections to the database server is (<i>max_connections</i> - <i>superuser_reserved_connections</i>). As a general rule, if you need more than 200 connections it is advisable to use connection pooling.</p><p><b>Available</b> - new connections allowed. </p><p><b>Used</b> - connections currently in use.</p>'
```
ilyam8, line 3913: Where do those changes come from? I don't see them in [netdata/netdata](https://github.com/netdata/netdata/blob/master/web/gui/dashboard_info.js).

jacekkolasa: @ilyam8 sorry for confusion, this file here is temporary, was used only for testing in the cloud. I'll remove it before this PR will be merged

ilyam8: Thanks for the clarification, @jacekkolasa 👍

burbuli8ra (approved):
In general it LGTM, but just a question, why was the **python version** section removed?

PR #441, opened 2022-08-31
fix: hash params without values

burbuli8ra:
This PR resolves issue #362 

#### Solution
* On hash (re)set, check for key/value pairs with `undefined` value before (re)assignment
* Fix eslintrc configuration
* Enable prettier by adding the dependency

https://user-images.githubusercontent.com/38426486/187630231-215a3923-14ed-48e1-a228-12a182cdf521.mov


src/utils/hash-utils.ts
```diff
@@ -28,7 +28,7 @@ export const makeHashFromObject = (params: { [paramKey: string]: string }) => {
     return ""
   }
   return entries
-    .map(([key, value]) => `${key}=${encodeURIComponent(value)}`)
+    .map(([key, value]) => value !== undefined ? `${key}=${encodeURIComponent(value)}` : key)
```
jacekkolasa, line 31: personally, i find removing unnecessary negations to be more readable, like:
```suggestion
    .map(([key, value]) => value === undefined ? key : `${key}=${encodeURIComponent(value)}`)
```
Which removes `not not` condition for `key`. But if for you it's more readable, it's ok :)

jacekkolasa (approved):
good catch!

AlexNti (approved):

PR #446, opened 2022-09-09
Added fallback for number chart to gauge.

jjtsou:
- When we change the mainheads to use number chart (mainheads in dashboard_info.js), then agent will be backwards compatible using gauge chart.


burbuli8ra (approved):

jacekkolasa:
@jjtsou dashboard.js is an old file, used only for accessing first version of dashboard (https://frankfurt.my-netdata.io/old/), done before migration to React.

PR #467, opened 2023-09-26
Allow building with modern NodeJS 

peat-psuwit:
The old `devDependencies` of this package causes 2 problems:
  - It makes building this package in the distributions difficult. Many distros no longer package NodeJS 12 that this project targeted.
  - The netdata/devenv package has its NodeJS upgraded to version 20 recently [1]. This breaks development of this package.

This PR aims to modernize the build-time dependencies so that it can be built with modern version of NodeJS (version 18 and newer). This will break support of NodeJS 12 this project is currently using. However, since NodeJS 12 is EOL'ed for over a year already, I think it's time to move on.

Updating the runtime dependencies of this project is _not_ the aim of this PR.

Note that due to behavior changes in NPM v7 and above [2], we need to use `--legacy-peer-deps` flag every time we install packages with NPM. This boils down to the conflict between `@netdata/react-filter-box`'s `dependencies` which contain `react@^16.6.3`, and `@netdata/netdata-ui`'s `peerDependencies` which contains `react@^18.2.0` (even in version `2.0.0` specified in `package-lock.json`). Because of this conflict, we can't just raise this package's `peerDependencies` of `react`, and will have to fix the `dependencies` of `@netdata/react-filter-box` first.

[1] https://github.com/netdata/community/pull/85
[2] https://stackoverflow.com/a/66620869


peat-psuwit:
At the moment I can get dependencies to install on NodeJS 20. I've upgraded react-scripts, Typescript to the latest versions, fixed various typing errors, and fixed Webpack errors. I'm still stuck with ESLint dependencies (I suspect I have to upgrade all of them & fix various errors that will appear), but I'm out of time now so I'll leave it here in case someone is interested in continuing the work. If I happen to find some time again, I might return to continue this work.

peat-psuwit:
I now manage to build this package under NodeJS 20 and NodeJS 18, and so I believe this is ready for review.

